PR #20, opened 2017-11-20
Fix Job.poll_

### LINKED ISSUE #14: Job.prototype.poll_ does not callback with err when job failed.
_From @ziplokk1 on November 15, 2017 20:57_

Description: 
Job.prototype.poll_ does not callback with err when job failed.
This is causing issues with Job.promise since the promise will always resolve instead of reject

Problem snippet:

```javascript
/**
 * Poll for a status update. Execute the callback:
 *
 *   - callback(err): Job failed
 *   - callback(): Job incomplete
 *   - callback(null, metadata): Job complete
 *
 * @private
 *
 * @param {function} callback
 */
Job.prototype.poll_ = function(callback) {
  this.getMetadata(function(err, metadata, apiResponse) {
    if (!err && apiResponse.status && apiResponse.status.errors) {
      // ******************************************
      // Here is where the issue lies.
      // For some reason err is undefined which is 
      // causing the following if statement condition to 
      // be false, thus not calling the callback with the error.
      // 
      // The solution is just putting `new` before common.util.ApiError.
      // err = new common.util.ApiError(apiResponse.status);
      err = common.util.ApiError(apiResponse.status);
    }
    if (err) {
      callback(err);
      return;
    }

    if (metadata.status.state !== 'DONE') {
      callback();
      return;
    }

    callback(null, metadata);
  });
};
```

Environment Details:
  - OS: agnostic
  - Node.js version: v6.11.2
  - npm version: 3.10.10
  - @google-cloud/bigquery version: 0.10.0

Steps to reproduce:

```javascript
const bigquery = require('@google-cloud/bigquery')();
// Query a table that doesn't exist to cause the job to fail.
bigquery.startQuery({query: "SELECT * FROM dataset.abc_this_table_doesnt_exist;"})
    .then(([job, apiResponse]) => {
        job.on('error', function (err) {
            // Never called
            console.error(err);
        });
        job.on('complete', function (meta) {
            console.log('success');
            console.log(meta.status.errors);
        });
    });
```

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2745_

ziplokk1:
ApiError was never being instantiated and thus the 'err' variable was always undefined. This caused issues because it would never errback.

This affected Job.on('error', ...) which in turn affected Job.promise, Table.copy, Table.copyFrom, Table.export, and Table.import.

Fixes #14

### DIFF
--- a/src/job.js
+++ b/src/job.js
@@ -378,7 +378,7 @@ Job.prototype.getQueryResultsStream = common.paginator.streamify(
 Job.prototype.poll_ = function(callback) {
   this.getMetadata(function(err, metadata, apiResponse) {
     if (!err && apiResponse.status && apiResponse.status.errors) {
-      err = common.util.ApiError(apiResponse.status);
+      err = new common.util.ApiError(apiResponse.status);
     }
 
     if (err) {


stephenplusplus:
Thanks again!

ziplokk1:
No problem!

PR #23, opened 2017-11-21
System Tests: Correct table notation.

stephenplusplus:
This issue popped up today:

```
ApiError: Syntax error: Unexpected "[" at [1:17]. If this is a table identifier, escape the name with `, e.g. `table.name` rather than [table.name].
```
\- https://circleci.com/gh/googleapis/nodejs-bigquery/183

I went through and corrected the locations where we were using the `[table]` pattern.


stephenplusplus:
@callmehiphop ping-o.

PR #29, opened 2017-12-08
Fix samples, upgrade repo-tools, and regenerate scaffolding.

jmdobry:


jmdobry:
dunno why lint fails here, it worked locally

stephenplusplus:
Probably using an old prettier. It's been changing regularly with new rules that break the builds.

PR #33, opened 2017-12-12
Samples: rewrite simple app tutorial.

tswast:
- Add additional region tags to support tutorial rewrite.
- Use more direct map for printing results.
- Use more relevant query from public datasets.

### DIFF
--- a/samples/queries.js
+++ b/samples/queries.js
@@ -20,41 +20,42 @@ function printResult(rows) {
   // [START bigquery_simple_app_print]
   console.log('Query Results:');
   rows.forEach(function(row) {
-    let str = '';
-    for (let key in row) {
-      if (str) {
-        str = `${str}\n`;
-      }
-      str = `${str}${key}: ${row[key]}`;
-    }
-    console.log(str);
+    let url = row['url'];
+    let viewCount = row['view_count'];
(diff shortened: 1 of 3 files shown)


tswast:
/cc @lukesneeringer @jmdobry 

(I don't have write access here, so can't add you as reviewers)

samples/queries.js
```diff
@@ -20,41 +20,42 @@ function printResult(rows) {
...
+  // [END bigquery_simple_app_client]
+
+  // [START bigquery_simple_app_query]
+  // The SQL query to run
+  const sqlQuery = `SELECT
+    CONCAT(
+      'https://stackoverflow.com/questions/',
+      CAST(id as STRING)) as url,
+    view_count
+    FROM \`bigquery-public-data.stackoverflow.posts_questions\`
```
stephenplusplus: The auto-linter (`npm run prettier`) is getting a bit tripped up over this line. Escaping seems the most natural solution, but feel free to suggest an alternative.

tswast: Escaping works for me.

tswast:
The docs use a commit hash to freeze the sample to a specific version, so this should be safe to merge even though I changed some of the region tags.

PR #42, opened 2017-12-19
rename job methods from start{Type} to create{Type}Job

callmehiphop:
- [x] Tests and linter pass

### DIFF
--- a/src/dataset.js
+++ b/src/dataset.js
@@ -222,6 +222,31 @@ function Dataset(bigQuery, id) {
 util.inherits(Dataset, common.ServiceObject);
 
 /**
+ * Start running a query scoped to your dataset.
+ *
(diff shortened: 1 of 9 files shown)


src/dataset.js
```diff
@@ -221,6 +221,31 @@ function Dataset(bigQuery, id) {
 
 util.inherits(Dataset, common.ServiceObject);
 
+/**
+ * Start running a query scoped to your dataset.
```
stephenplusplus: We should copy the gist of the parent method's description:

 * Run a query as a job. No results are immediately returned. Instead, your
 * callback will be executed with a {@link Job} object that you must
 * ping for the results. See the Job documentation for explanations of how to
 * check on the status of the job.

Then link to the upstream.

callmehiphop: Done - PTAL!

PR #46, opened 2017-12-23
more description for nextQuery

alexander-fenster:
Please check if the update is valid.

### DIFF
--- a/src/job.js
+++ b/src/job.js
@@ -285,6 +285,8 @@ Job.prototype.cancel = function(callback) {
  * @param {array} rows The results of the job.
  * @param {?object} nextQuery A pre-made configuration object for your next
  *     request. This will be `null` if no additional results are available.
+ *     If the query is not yet complete, you may get empty `rows` and 
+ *     non-`null` `nextQuery` that you should use for your next request.
  * @param {object} apiResponse The full API response.
  */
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Is this true? I would think the API returns some, but not all, rows in each API response.

alexander-fenster:
@stephenplusplus That was the suggested edit by the API owners. As I understand, what it's supposed to say is that if the query has not started returning any rows–i.e. is incomplete,–the rows array might be empty, with non-null nextQuery. Do you have any suggestions how it can be written better?

stephenplusplus:
I guess the question would be if the [getQueryResults API](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults) ever returns `pageToken` but an empty `rows`. I didn't think that situation could occur, but if the API team confirms that it can, then the edits in this PR are perfect.

stephenplusplus:
Looks like there's another property in the picture, `jobComplete`:

> **jobComplete** If rows or totalRows are present, this will always be true

So, it sounds like rows could be empty and `jobComplete` can be false, meaning we'll execute the callback with an empty array for rows, just like the API team said :) Sorry for the delay. Let's merge this thing.

PR #60, opened 2018-02-14
Update examples that had bigquery.table

polds:
bigquery has no method table, but dataset does.

Fixes internal bug


stephenplusplus:
Thank you!

PR #65, opened 2018-03-05
Documentation update for the change from `import` to `loaded`

snird:
One link linked to the old not-existing `Table#import` method, it had to be updated to `Table#load`.
Also updated some places saying the file and data is "imported" instead of "loaded", to be more consistent with new terminology and less confusing.


stephenplusplus:
Thanks! After the CLA is figured out, I will merge asap!

snird:
signed it. :+1:

stephenplusplus:
Awesome, thanks again!

PR #66, opened 2018-03-05
Add samples for loading CSV from GCS

tswast:
These samples are intended to appear in the documentation at https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-csv

See the samples at https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json for similar work in other languages.

- [x] Tests and linter pass - I couldn't get even the existing tests to pass. It seems `ava` is not getting any output from the samples sometimes. (Maybe it is not waiting for the sample to finish running?)


samples/tables.js
```diff
@@ -304,6 +304,301 @@ function loadFileFromGCS(datasetId, tableId, bucketName, filename, projectId) {
...
+   */
+  const bucketName = "cloud-samples-data";
+  const filename = "bigquery/us-states/us-states.csv";
+
+  // Instantiates clients
+  const bigquery = new BigQuery({
+    projectId: projectId,
+  });
+
+  const storage = Storage({
```
jmdobry: s/`= Storage({`/`= new Storage({`

samples/tables.js
```diff
@@ -304,6 +304,301 @@ function loadFileFromGCS(datasetId, tableId, bucketName, filename, projectId) {
...
+  // const datasetId = "my_dataset";
+  // const tableId = "my_table";
+
+  /**
+   * This sample loads the CSV file at
+   * https://storage.googleapis.com/cloud-samples-data/bigquery/us-states/us-states.json
+   *
+   * TODO(developer): Replace the following lines with the path to your file.
+   */
+  const bucketName = "cloud-samples-data";
```
jmdobry: Use single quotes throughout.

samples/tables.js
```diff
@@ -304,6 +304,301 @@ function loadFileFromGCS(datasetId, tableId, bucketName, filename, projectId) {
...
+  });
+
+  const storage = Storage({
+    projectId: projectId,
+  });
+
+  // Configure the load job. For full list of options, see:
+  // https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load
+  const metadata = {
+    'sourceFormat': 'CSV',
```
jmdobry: Remove single quotes from the key names.

samples/tables.js
```diff
@@ -304,6 +304,301 @@ function loadFileFromGCS(datasetId, tableId, bucketName, filename, projectId) {
...
+  bigquery
+    .dataset(datasetId)
+    .table(tableId)
+    .load(storage.bucket(bucketName).file(filename), metadata)
+    .then(results => {
+      job = results[0];
+      console.log(`Job ${job.id} started.`);
+
+      // Wait for the job to finish
+      return job;
```
jmdobry: > `return job;`

This doesn't seem right to me, since when has just `return job;` worked? Looking at https://github.com/googleapis/nodejs-common/blob/master/src/operation.js#L94, it would seem this should be `return job.promise;`. That would mean the existing bigquery samples in this file are wrong too.

tswast: Seems to make the tests actually wait for the job now, but now they are failing for another reason.

```
$ node tables.js load-gcs-csv swast-scratch test_dataset us_states_csv
Job swast-scratch:US.5d1c5cc9-1ac1-436e-8d14-3c59f60a58fe started.
ERROR: TypeError: Cannot read property 'status' of undefined
    at bigquery.dataset.table.load.then.then.metadata (/Users/swast/src/nodejs-bigquery/samples/tables.js:367:31)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

Based on your link, it does seem the metadata is supposed to be passed in from the promise, but apparently it's not.

stephenplusplus: I think we need to return `job.promise()`

tswast: Hmm... `ERROR: TypeError: job.promise is not a function`

stephenplusplus: Woops, looks like `load()` doesn't return an operation or Job object at all. When the callback or promise is resolved, the job is complete: https://github.com/googleapis/nodejs-bigquery/blob/fbe4f93f9167e904ee9d0beaf27f0d0b9da6d7bb/src/table.js#L1688

tswast: Good catch @stephenplusplus ! I've mailed https://github.com/googleapis/nodejs-bigquery/pull/67 to fix the existing samples.

samples/tables.js
tswast: Done.

tswast:
Note: I removed the assert on the number of rows after the `insertAll` (streaming) call, because that is flakey. Rows first go to the streaming buffer, where they don't appear in row listings.

tswast:
Ready to merge?

tswast:
Published! https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-csv

PR #81, opened 2018-03-29
feat(job): regionalization support

callmehiphop:
- [ ] Tests and linter pass
- [ ] Get `Job#cancel` to work.

Please take a look at the updated system tests for the usage!


system-test/bigquery.js
```diff
@@ -505,6 +505,115 @@ describe('BigQuery', function() {
...
+
+      it.skip('should cancel a job', function() {
+        return dataset.createQueryJob(QUERY).then(function(data) {
+          var job = data[0];
+
+          return job.cancel();
+        });
+      });
+
+      it('should get query results', function() {
```
tswast: I'd like to see some tests that show that a query, copy, extract fail if the job location doesn't match the table location. (For example, try to query with job location = 'US' but this table is in the Asia region)

tswast: See https://github.com/GoogleCloudPlatform/google-cloud-python/blob/d44dc25bb05f94c71dba0beef32c5303eaa38641/bigquery/tests/system.py#L610-L706 for an example.

callmehiphop:
@tswast @jjolma I'm having some difficulties cancelling jobs when a location is present. When I trace my request this is what I get

```js
{
  uri: 'https://www.googleapis.com/bigquery/v2/project/boreal-physics-911/jobs/bd31d376-f86f-4488-bef0-6b94fcacc243/cancel?location=asia-northeast1',
  method: 'POST',
  headers: {
    'User-Agent': 'gcloud-node-bigquery/1.0.0',
    'x-goog-api-client': 'gl-node/8.10.0 gccl/1.0.0',
    Authorization: 'Bearer xxxx',
    host: 'www.googleapis.com',
    'accept-encoding': 'gzip, deflate',
    'content-length': 0
  }
}
```

Which results in the following response

```js
{
  "error": {
    "errors": [
      {
        "domain": "global",
        "reason": "notFound",
        "message": "Not found: Job boreal-physics-911:bd31d376-f86f-4488-bef0-6b94fcacc243"
      }
    ],
    "code": 404,
    "message": "Not found: Job boreal-physics-911:bd31d376-f86f-4488-bef0-6b94fcacc243"
  }
}
```

Does anything look obviously wrong here?

callmehiphop:
@stephenplusplus @alexander-fenster this should be ready for a review, if either of you would be so kind!

src/index.js
```diff
@@ -895,10 +909,17 @@ BigQuery.prototype.createJob = function(options, callback) {
...
+          errors: resp.status.errors,
+          response: resp,
+        });
+      }
+
       var job = self.job(jobId);
       job.metadata = resp;
 
-      callback(null, job, resp);
+      callback(err, job, resp);
```
stephenplusplus: So we should still return a job even though we returned an error?

src/job.js
```diff
@@ -236,6 +214,9 @@ util.inherits(Job, common.Operation);
  *
  * @see [Jobs: get API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/jobs/cancel}
  *
+ * @param {object} [options] Configuration object.
+ * @param {string} [options.location] The geographic location of the job
```
stephenplusplus: Needs a trailing `.`

src/index.js
callmehiphop: I think I want some one on the BQ team to weigh in here (@tswast @jjolma). I'm not sure if an error is present in the job status that it means the job should be discarded.

system-test/bigquery.js
```diff
@@ -505,6 +505,185 @@ describe('BigQuery', function() {
...
+      before(function() {
+        // create a dataset in a certain location will cascade the location
+        // to any jobs created through it
+        return dataset.create({location: LOCATION}).then(function() {
+          return table.create({schema: SCHEMA});
+        });
+      });
+
+      it('should create a load job', function() {
+        return table.createLoadJob(TEST_DATA_FILE).then(function(data) {
```
stephenplusplus: This should be done in the before so that the other tests don't depend on this test being run and succeeding. This test can either be skipped, or it can be something like a check on the metadata to be sure the name/location/metadata was set properly.

system-test/bigquery.js
```diff
@@ -505,6 +505,185 @@ describe('BigQuery', function() {
...
+          badJob.location = 'US';
+
+          badJob.cancel(function(err) {
+            assert.strictEqual(err.code, 404);
+            done();
+          });
+        });
+
+        it('should cancel a job', function(done) {
+          job.cancel(function(err) {
```
stephenplusplus: job.cancel(done)

system-test/bigquery.js
```diff
@@ -505,6 +505,185 @@ describe('BigQuery', function() {
...
+        before(function() {
+          return dataset.createQueryJob(QUERY).then(function(data) {
+            job = data[0];
+          });
+        });
+
+        it('should fail if the job location is incorrect', function(done) {
+          var badJob = bigquery.job(job.id);
+
+          badJob.location = 'US';
```
stephenplusplus: I think having a dedicated `setLocation()` could be more clear. Prior art: [`file.setEncryptionKey()`](https://github.com/googleapis/nodejs-storage/blob/fb88bd16b25e80962e630a300632d0d15817682c/src/file.js#L1251). To pass it through during instantiation: `bucket.file({ encryptionKey: '...' })`

Do we have other places where we want the user to set a property on a Service/ServiceObject?

src/index.js
tswast: Yes, the job should be available even on errors. There are sometimes some useful info for debugging in the `statistics` property of the job.

system-test/bigquery.js
callmehiphop: Not that I'm aware of. I guess I don't see a huge difference between making a dedicated setter vs. allowing the user to configure a public property, however I can definitely get on board with allowing the user to specify the location during instantiation.

stephenplusplus: I like having the setter for consistency now because of Storage, but also for documentation. Having a method with examples and its own docs is much easier to notice. Updating a property doesn’t exactly have the same effect, e.g. in Duplexify you call “setWritable” as opposed to updating a “writable” property.

callmehiphop: That's true, but I'm not sure if our users will benefit from or ever need to set the language asynchronously. IMO I think passing it in during instantiation is the best way to go.

src/job.js
callmehiphop: Fixed!

system-test/bigquery.js
callmehiphop: Done!

system-test/bigquery.js
callmehiphop: Done!

system-test/bigquery.js
callmehiphop: I made a commit to allow passing the location in via constructors. PTAL!

src/dataset.js
```diff
@@ -30,13 +30,16 @@ var Table = require('./table.js');
  * @class
  * @param {BigQuery} bigQuery {@link BigQuery} instance.
  * @param {string} id The ID of the Dataset.
+ * @param {object} [metadata] Dataset metadata.
+ * @param {string} [metadata.location] The geographic location of the dataset.
+ *      Required except for US and EU.
```
jjolma: Can you replace the this line with "Defaults to US."?

src/dataset.js
```diff
@@ -554,6 +597,9 @@ Dataset.prototype.query = function(options, callback) {
  * Create a Table object.
  *
  * @param {string} id The ID of the table.
+ * @param {object} [metadata] Table metadata.
+ * @param {string} [metadata.location] The geographic location of the table.
+ *      Required except for US and EU.
```
jjolma: Can you remove the line about it being required? I'd suggest what we have in our discovery doc: The geographic location where the table resides. This value is inherited from the dataset.

jjolma: Also, can you make it clear that the table's location is read-only? This library should populate it from API responses, but callers should not populate it.

src/dataset.js
callmehiphop: Done!

system-test/bigquery.js
```diff
@@ -505,6 +505,180 @@ describe('BigQuery', function() {
...
+            return table.createCopyJob(otherTable).then(function(data) {
+              var job = data[0];
+
+              assert.strictEqual(job.location, LOCATION);
+            });
+          });
+        });
+
+        describe('extract', function() {
+          var extractFile = bucket.file('location-extract-data.json');
```
tswast: I believe this bucket needs to be in the same region as the dataset.

src/dataset.js
callmehiphop: Absolutely! I just pushed a commit with some updated verbiage

system-test/bigquery.js
stephenplusplus: So the only way we want the user to set the location is through the constructors, right? Let's remove the getters/setters that allow `{object}.location = '...'` flow. That leaves a backdoor of undocumented behavior that will be confusing to maintain.

callmehiphop: Okie dokie, I just pushed a commit that removes it.

src/dataset.js
```diff
@@ -205,6 +208,22 @@ function Dataset(bigQuery, id) {
...
+   * @name Dataset#metadata
+   * @type {object}
+   */
+  this.metadata = metadata || {};
+
+  /**
+   * @name Dataset#location
+   * @type {string}
+   */
+  Object.defineProperty(this, 'location', {
```
stephenplusplus: This pattern is unusual to me. metadata could be considered a protected property that is only supposed to be a reflection of the remote object. Can we just localalize the "location" property of a user-provided options object, and then embed it later as part of whatever requests need it?

callmehiphop: Well, the issue I was facing was that if the location comes back through any `apiResponse` we should be using it, so in some cases the user might not even provide a location.

stephenplusplus: Ah, gotcha. And if we didn't include it in follow up responses, it would error? If that's the case, it makes total sense to me! Could you just add a little inline comment to explain that?

callmehiphop: Unfortunately I won't be able to find the time to do that today. :(

callmehiphop: Ok, I've added a developer comment explaining this. PTAL

src/dataset.js
```diff
@@ -30,13 +30,16 @@ var Table = require('./table.js');
  * @class
  * @param {BigQuery} bigQuery {@link BigQuery} instance.
  * @param {string} id The ID of the Dataset.
+ * @param {object} [metadata] Dataset metadata.
```
stephenplusplus: We should rename this to `options` and `options.location`. It shouldn't have a connection to "Dataset metadata" from the user's perspective, because that indicates it has a connection with the remote, upstream object. For example, the user could provide other Dataset metadata, and assume that it will be used in API responses and update that value. We just want the user to provide a `location`, if they need to, through a configuration object, like we usually do;

```js
* @param {object} [options] Configuration object.
* @param {string} [options.location] The geographic...
```

src/dataset.js
```diff
@@ -240,11 +265,18 @@ Dataset.prototype.createQueryJob = function(options, callback) {
     };
   }
 
-  options = extend(true, {}, options, {
-    defaultDataset: {
-      datasetId: this.id,
+  options = extend(
+    true,
+    {
+      location: this.location,
```
stephenplusplus: Can we avoid having this option, and just let the user either provide the location from the `bigquery.dataset()` constructor, or have it set already for the objects we give to the user?

From the req sheet:

> Allow the user to optionally specify a location for job operations.  Consider having one flag that can be used for dataset creation (the existing location field) and the job operations.

This same question applies to the other objects and methods as well.

src/dataset.js
```diff
@@ -554,6 +600,13 @@ Dataset.prototype.query = function(options, callback) {
  * Create a Table object.
  *
  * @param {string} id The ID of the table.
+ * @param {object} [metadata] Table metadata.
```
stephenplusplus: Same comment about `[metadata] Table metadata.` ➡️ `[options] Configuration object.`

src/index.js
```diff
@@ -907,15 +928,18 @@ BigQuery.prototype.createJob = function(options, callback) {
  * Create a reference to a dataset.
  *
  * @param {string} id ID of the dataset.
+ * @param {object} [metadata] Dataset metadata.
```
stephenplusplus: Same comment about `[metadata] Dataset metadata.` ➡️ `[options] Configuration object.`

src/dataset.js
callmehiphop: I've reviewed some of the other languages implementations and there is similar functionality like this in the other clients. That being said, I think we should keep this.

src/dataset.js
callmehiphop: Done.

src/dataset.js
callmehiphop: Done.

src/index.js
callmehiphop: Done.

src/dataset.js
stephenplusplus: cc @tswast @jjolma 

For Node, if a property needs to be changed, we recommend the user creates a new instance, e.g.

```js
const dataset1 = bigquery.dataset('name', { location: 'location-1' })
const dataset2 = bigquery.dataset('name', { location: 'location-2' })
```

Then, all operations on that object are correctly scoped:

```js
dataset1.createQueryJob(...) // uses `location-1`
dataset2.createQueryJob(...) // uses `location-2`
```

This is similar to how the user can use different project IDs, auth credentials, etc:

```js
const BigQuery = require('@google-cloud/bigquery')
const bigQueryProject1 = new BigQuery({ project: 'project-id-1' })
const bigQueryProject2 = new BigQuery({ project: 'project-id-2' })
```

stephenplusplus: Correct, I think we should only support passing into the constructor. It maintains the correct hierarchy, as the pairing of a `location` to a Dataset is similar to the pairing of a `name`, in that they both need to be known up front. Lazy-setting of the location feels similar to lazy-setting a name. In other words, we don't allow the following pattern, so I don't see why we should allow it for location:

```js
const dataset = bigquery.dataset();
const [metadata] = await dataset.getMetadata({ name: 'dataset-name' });
```

callmehiphop: I'm not sure I agree with that stance, `location` being a query parameter, as opposed to a path parameter, makes me think it's ok to configure it in this fashion. I'm not opposed to removing it, but I'd prefer to hear from the BigQuery team before doing so, just in case they think there is some value in it.

tswast: Yes, location is something that needs to be set before the API call to create the dataset. In an ideal world it would have been a path parameter, but such a change wouldn't be backwards compatible.

Note: the [table encryption key](https://cloud.google.com/bigquery/docs/customer-managed-encryption) is similar to this in that it can't be set if you create a table without one (though unlike location, it can be changed to a new value after the fact)

callmehiphop: Very well! I've went ahead and removed this. @stephenplusplus PTAL

src/dataset.js
```diff
@@ -205,6 +208,28 @@ function Dataset(bigQuery, id) {
 
   this.bigQuery = bigQuery;
 
+  /**
+   * @name Dataset#metadata
+   * @type {object}
+   */
+  this.metadata = options || {};
```
stephenplusplus: Sorry to keep bringing this up. I think we need to completely separate `options` from `metadata`:

```js
this.metadata = {};

if (options.location) {
  this.metadata.location = options.location;
}
```

src/index.js
```diff
@@ -1196,6 +1223,8 @@ BigQuery.prototype.job = function(id) {
  * @param {string|object} query A string SQL query or configuration object.
  *     For all available options, see
  *     [Jobs: query request body](https://cloud.google.com/bigquery/docs/reference/v2/jobs/query#request-body).
+ * @param {string} [query.location] The geographic location of the job.
```
stephenplusplus: Should `location` be an argument to the BigQuery constructor as well, that would be used for this operation, and descend to the other levels?

src/dataset.js
```diff
@@ -205,6 +208,28 @@ function Dataset(bigQuery, id) {
...
+
+  /*!
```
stephenplusplus: Now that we're not allowing lazy-setting of `location`, what are the scenarios that this lazy-getting would be used?

src/job.js
```diff
@@ -135,37 +138,6 @@ function Job(bigQuery, id) {
      */
     get: true,
 
-    /**
```
stephenplusplus: *The following comment is only valid if we don't need to support lazy-getting anymore*

This can be reverted, like this:

```js
getMetadata: {
  qs: {location: this.location}
}
```

src/table.js
```diff
@@ -846,6 +901,10 @@ Table.prototype.createExtractJob = function(destination, options, callback) {
     delete options.jobPrefix;
   }
 
+  if (this.location) {
```
stephenplusplus: Why is `options.location` not accepted here, but it is in, e.g. `createCopyFromJob()`?

src/index.js
callmehiphop: sure

src/dataset.js
callmehiphop: I found this to be conveinent when building requests so I didn't have to do something like this all over the place.

```js
var qs = {location: this.metadata && this.metadata.location};

// or for job requests

if (this.metadata && this.metadata.jobReference) {
  qs = {location: this.metadata.jobReference.location};
}
```

Because the location would be in different spots depending on the object, normalizing the location seems like a pretty convenient thing to do for the user on the off chance they wanted to reference it.

src/job.js
callmehiphop: This actually seems slightly less efficient to me, if there is no location then I see no reason to create an object here. I realize the difference would be negligible, but if its all the same I see no reason to change it.

src/dataset.js
stephenplusplus: I'm cool with the aliasing, but if we're always getting the value up front from the user, couldn't we just do:

```js
this.location = options.location;
```

The comment talks about how we cache it if the location comes back in an apiResponse, but I think we're not relying on that anymore.

src/job.js
stephenplusplus: In this PR, you had to re-create the whole `getMetadata()` function, whereas before, we inherit it from ServiceObject. It looks like you were forced to do that, because we needed to provide a `qs` object, but `ServiceObject` actually has this scenario in mind, and allows providing the `qs` object up front during instantiation.

Quick correction on my part, though:

```js
getMetadata: {
  reqOpts: {
    qs: {location: this.location}
  }
}
```

src/dataset.js
callmehiphop: Ah, I understand. I think there's still a need for this to allow the user to do something like so

```js
const dataset = bigquery.dataset('name');

await dataset.get();
const [job] = await dataset.createQueryJob({...});
```

src/job.js
callmehiphop: Oh sorry, I misunderstood in my earlier comment. I think we want to continue with the lazy getting approach as mentioned previously, so this can't be reverted.

src/dataset.js
callmehiphop: Done.

src/index.js
callmehiphop: Done.

src/dataset.js
stephenplusplus: Per the other conversation about hierarchy, I don't think we should support this behavior, since we demonstrate in the examples and documentation that `location` should be set on the instance.

The user would have to accidentally fall into the pattern in your example above, unless there's a utility in fetching the location of a dataset from just a name.

But that has me wondering-- wouldn't `dataset.get()` fail if `name` is in a location other than the upstream API's default location? (sub-question-- can the same dataset name be used in two locations?)

callmehiphop: You're right, it would fail. Using whatever location came back in the `apiResponse` was in the pre-reqs, so I would prefer that @tswast or @jjolma make this decision.

src/index.js
```diff
@@ -1183,8 +1230,8 @@ BigQuery.prototype.getJobsStream = common.paginator.streamify('getJobs');
  *
  * const myExistingJob = bigquery.job('job-id');
  */
-BigQuery.prototype.job = function(id) {
-  return new Job(this, id);
+BigQuery.prototype.job = function(id, options) {
+  return new Job(this, id, options);
```
stephenplusplus: Should we pick up `this.location` here?

src/table.js
stephenplusplus: `createCopyJob` still accepts `metadata.location`. Also, `metadata.location` is documented for `copy`, `copyFrom`, and `createCopyJob`. I'm only mentioning it in this thread and not as line comments, because I'm not sure if I'm just confused and they are correct, or if they need to be removed.

callmehiphop: They needed to be removed. Which I have done

src/index.js
callmehiphop: Yes, done.

callmehiphop:
@stephenplusplus is there anything else that needs to be done here?

src/dataset.js
stephenplusplus: I went back to read the doc:

> - Extract the job’s location from any responses containing a JobReference.
> - Include a job’s location for requests operating on a fetched job. For example, if the client does a jobs.list and then does a jobs.cancel on one of the jobs, the cancel should include the job’s location.

For the methods that return a `JobReference`, we don't need to extract the location, because the user provided the location up-front. The only case I can think of (correct me if I'm wrong), where we don't know the location is `getJobs()`. Thus, I believe we can accomplish both requirements by changing `getJobs()`:

```js
var jobs = (resp.jobs || []).map(function(jobObject) {
  var job = that.job(jobObject.jobReference.jobId, {
    location: jobObject.jobReference.location,
  });
  job.metadata = jobObject;
  return job;
});
```

stephenplusplus:
I think we're still waiting on some clarity for https://github.com/googleapis/nodejs-bigquery/pull/81#discussion_r178682971

src/dataset.js
callmehiphop: I suppose that's fair, we'd probably want to do the same thing for `getDatasets` as well. However I think I'm struggling to understand why this is superior to lazy getting. I could be wrong, but this feels like bike shedding.

stephenplusplus: Since we can avoid it, the lazy-getting creates a confusing point in the code, that doesn't have a purpose. Leaving leaks of hidden behavior and support for un-documented use cases makes the code hard to maintain. I just figure we should make the change as deliberate and simple as possible, so as to make it easier to understand/refactor some day.

PR #85, opened 2018-04-02
wait for job result before emitting complete.

### LINKED ISSUE #8: bigquery:createWriteStream() should propogate insertErrors
_From @chrishiestand on September 29, 2017 4:30_

The gist is bigquery sometimes silently drops one or more streams when many streams are used in parallel (42 in this example). If I'm hitting a quota an error ought to be thrown, but no errors are thrown the program runs as expected but in the end there are rows missing from the bigquery data.

If there is a bug, it might be in gcloud-node, or it might be in the bigquery api. Both seem less likely than me having made a mistake, so I hope you can find something I've done wrong.

The tricky part is the bug doesn't always reproduce. When the bug does reproduce `n` streams of size `s` are dropped so bigquery will have `n` * `s` missing rows. So if `n=2` and `s=150` bigquery will be missing 300 rows. In other words, the problem does not appear to be that a subset of stream data is missing, but rather one or more entire streams are missing.

This seems to reproduce reliably sometimes, and other times it reliably does not reproduce. To try and get the opposite result, try again later and/or change the stream load with the env variables.

This small bug reproduction project <https://github.com/chrishiestand/gcloud-node-bigquery-manystreams-bug>is the result of troubleshooting missing big query data in a production system where 50 streams are processed in parallel.

Below is a screenshot of the reproduction repository showing a reproduction of the issue.

image

In contrast, if I reduce the number of streams from 42 to 10, the tests pass as below:
image

#### Environment details

  - OS: OS X 10.12.6
  - Node.js version: 8.6.0
  - npm version: 5.3.0
  - google-cloud-node version: @google-cloud/bigquery = 0.9.6

#### Steps to reproduce

Please go here for a reproduction project: <https://github.com/chrishiestand/gcloud-node-bigquery-manystreams-bug>

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2635_

stephenplusplus:
Fixes #8 

**This would be a breaking change**

Currently, if you call `table.createWriteStream()`, the stream will emit "complete" with a Job object. You can then use that Job object to assign event handlers, as usual with an LRO. Judging from #8, I'm not sure this pattern is clear to the users. And aside from that, I'm not sure what we're doing is ideal, either.

Ideally, we would not emit "complete" until the Job finished, instead of making the user do it manually. This PR shows how easy it would be for us to implement. However, I *believe* this would be a breaking change, although it's hard to say if it's actually a bugfix.

I'm leaving this open for discussion on how we should handle this. Please share your thoughts!

---

### Before
```js
table.createWriteStream(...)
  .on('complete', job => {
    job
      .on('error', err => {
        // Job failed.
      })
      .on('complete', () => {
        // Job completed successfully.
      });
  });
```

### After
```js
table.createWriteStream(...)
  .on('error', err => {
    // Job failed.
  })
  .on('job', job => {
    // `job` is a Job object.
  })
  .on('finish', () => {
    // Job completed successfully.
  });
```

And, `complete` still works to combine those. I added the "finish" event, because that's the natural writable stream closing event.

```js
table.createWriteStream(...)
  .on('error', err => {
    // Job failed.
  })
  .on('complete', job => {
    // `job` is a Job object.
    // Job completed successfully.
  });
```

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -1231,7 +1231,10 @@ Table.prototype.createWriteStream = function(metadata) {
         });
 
         job.metadata = data;
-        dup.emit('complete', job);
+
+        job
+          .on('error', err => dup.destroy(err))
+          .on('complete', () => dup.emit('complete', job));
       }
     );
   });
--- a/src/table.ts
+++ b/src/table.ts
@@ -1212,6 +1212,11 @@ Table.prototype.createWriteStream = function(metadata) {
 
   const dup = streamEvents(duplexify());
 
+  // This will allow the natural `finish` event to fire.
+  dup.on('prefinish', () => {
+    dup.cork();
+  });
+
   dup.once('writing', function() {
     common.util.makeWritableStream(
       dup,
@@ -1245,7 +1250,10 @@ Table.prototype.createWriteStream = function(metadata) {
 
         job
           .on('error', err => dup.destroy(err))
-          .on('complete', () => dup.emit('complete', job));
+          .on('complete', () => {
+            dup.emit('complete', job);
+            dup.uncork();
+          });
       }
     );
   });
--- a/src/table.ts
+++ b/src/table.ts
@@ -1142,9 +1142,12 @@ Table.prototype.createReadStream = paginator.streamify('getRows');
  *
  * request.get(csvUrl)
  *   .pipe(table.createWriteStream(metadata))
- *   .on('complete', function(job) {
+ *   .on('job', function(job) {
  *     // `job` is a Job object that can be used to check the status of the
  *     // request.
+ *   })
+ *   .on('complete', function(job) {
+ *     // The job has completed successfully.
  *   });
  *
  * //-
@@ -1154,7 +1157,13 @@ Table.prototype.createReadStream = paginator.streamify('getRows');
  *
  * fs.createReadStream('./test/testdata/testfile.json')
  *   .pipe(table.createWriteStream('json'))
- *   .on('complete', function(job) {});
+ *   .on('job', function(job) {
+ *     // `job` is a Job object that can be used to check the status of the
+ *     // request.
+ *   })
+ *   .on('complete', function(job) {
+ *     // The job has completed successfully.
+ *   });
  */
 Table.prototype.createWriteStream = function(metadata) {
   const self = this;
@@ -1231,6 +1240,8 @@ Table.prototype.createWriteStream = function(metadata) {
         });
 
         job.metadata = data;
+
+        dup.emit('job', job);
 
         job
           .on('error', err => dup.destroy(err))
--- a/system-test/bigquery.ts
+++ b/system-test/bigquery.ts
@@ -740,10 +740,16 @@ describe('BigQuery', function() {
     });
 
     it('should insert rows via stream', function(done) {
+      let job;
+
       fs.createReadStream(TEST_DATA_JSON_PATH)
         .pipe(table.createWriteStream('json'))
         .on('error', done)
-        .on('complete', function() {
+        .on('complete', function(_job) {
+          job = _job;
+        })
+        .on('finish', function() {
+          assert.strictEqual(job.metadata.status.state, 'DONE');
           done();
         });
     });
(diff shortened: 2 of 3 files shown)


callmehiphop:
I'm not sure what the official rules are for making breaking changes are now that we're `1.x.x`. But we're definitely going to want a lower level equivalent that returns the job. If making a breaking change is too cumbersome, you could just create a new method with a different name that abstracts the job away.

stephenplusplus:
We could add an option, e.g. `waitForJobCompletion`, which would be false by default.

callmehiphop:
> We could add an option, e.g. waitForJobCompletion, which would be false by default.

The BigQuery team has previously requested that we provide both a high and low level method for all methods that create a job. Aside from that using an option to change the signature of a callback has generally been frowned upon in the past. Not to mention this behavior might be found confusing since no other job related methods follow this pattern.

stephenplusplus:
API team feedback is welcome. We wouldn't change the signature-- we can still emit "complete" with the job. IMO, we'd be making the behavior of the method more logical.

> Not to mention this behavior might be found confusing since no other job related methods follow this pattern.

Do we have stream methods that follow that pattern?

callmehiphop:
>  We wouldn't change the signature-- we can still emit "complete" with the job.

But the Job itself would be in a different state, hence why this is a breaking change. IMO that is a signature change.

I'll go with whatever the group decides, but my vote would be for a new higher level method instead of using an option.

stephenplusplus:
I think there's room for multiple interpretations of `breaking`-- to me, the contestants for considering this a breaking change are:

1) Waiting for a job to complete takes longer than before (code doesn't change, though)
2) We could emit an error where we didn't before (code doesn't change, but where there wasn't an error before, there now would be)

The candidates for this to be a bug fix to me are:

1) As discovered in the issue, it's not clear to users they have to assign further event handlers to the Job object (frankly, I didn't know, either)
2) Users could be getting the "complete" event and proceeding as if all operations succeeded, when in fact, there were errors

If we decide a breaking change, that's fine. When considering how to solve, I'll just mention that `createWriteStream` is a bit different than a regular method named after an upstream one, in that this uses the de facto writable-stream language with "createWriteStream" (e.g. `fs.createWriteStream`). I would prefer to solve with an option, so we can retain the logical name.

JustinBeckwith:
@stephenplusplus what's the status on this one?

JustinBeckwith:
I'm going to be honest - I'm not deep enough with this module to make an informed decision on what we should do :) We have a semver major bump coming up very soon, so if we're going to make a breaking change, now's the time.

I would trust the opinion @tswast here.

tswast:
I've definitely found it confusing that `complete` is emitted before the job actually finishes. I'm in favor of this change.

> The BigQuery team has previously requested that we provide both a high and low level method for all methods that create a job. Aside from that using an option to change the signature of a callback has generally been frowned upon in the past. Not to mention this behavior might be found confusing since no other job related methods follow this pattern.

We actually somewhat ignored this advice in the Python client because it made a whole lot more sense to make a flexible method for creating a job (well, technically we have a method for each job type) and provide multiple ways to wait for it to finish (blocking + asynchronously). I think the same logic applies to Node.js: we should use the the built-in ways to wait for things.

I'm happy so long as the following are satisfied:

* The initial function call to create the job is non-blocking,
* You can pass in a full job configuration to customize the job.
* The initial function call returns the initial job state (so that the job ID can be serialized / shared with the user before waiting),
* It's easy to wait for the job to complete.

stephenplusplus:
👍 

I made the change. Here's the old and new:

### Before
```js
table.createWriteStream(...)
  .on('complete', job => {
    job
      .on('error', err => {
        // Job failed.
      })
      .on('complete', () => {
        // Job completed successfully.
      });
  });
```

### After
```js
table.createWriteStream(...)
  .on('error', err => {
    // Job failed.
  })
  .on('job', job => {
    // `job` is a Job object.
  })
  .on('finish', () => {
    // Job completed successfully.
  });
```

And, `complete` still works to combine those. I added the "finish" event, because that's the natural writable stream closing event.

```js
table.createWriteStream(...)
  .on('error', err => {
    // Job failed.
  })
  .on('complete', job => {
    // `job` is a Job object.
    // Job completed successfully.
  });
```

PR #88, opened 2018-04-04
feat(job): support custom ids

### LINKED ISSUE #86: Support user-provided job ID for all job types
I got the following feedback from a user in the docs. https://cloud.google.com/bigquery/docs/running-jobs

> include a disclaimer here for nodejs ... no matter what we assign to this in a configuration resource object, the returned jobReference.jobId is always one created by the service, not the one supplied.

I believe all other languages have a `jobId` and `jobPrefix` parameter for all job types, which allows using a manually-specified job ID or a prefix for a randomly-generated job ID.

callmehiphop:
Fixes #86

- [x] Tests and linter pass

### DIFF
--- a/system-test/bigquery.js
+++ b/system-test/bigquery.js
@@ -286,6 +286,18 @@ describe('BigQuery', function() {
     });
   });
 
+  it('should honor the job id option', function(done) {
+    var jobId = 'hi-im-a-job-id';
+    var options = {query, jobId};
+
+    bigquery.createQueryJob(options, function(err, job) {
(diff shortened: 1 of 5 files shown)


system-test/bigquery.js
```diff
@@ -286,6 +286,18 @@ describe('BigQuery', function() {
     });
   });
 
+  it('should honor the job id option', function(done) {
+    var jobId = 'hi-im-a-job-id';
```
stephenplusplus: hi!

PR #93, opened 2018-04-13
Fixes broken image link to cloud shell button

alixhami:


jmdobry:
This is already fixed in nodejs-repo-tools. My recommendation is to re-run `npm run generate-scaffolding` at the root of this repo.

also, `//:` doesn't work for markdown links, we already tried it.

PR #94, opened 2018-04-13
Updates region tags to standard and deletes unused region tags

alixhami:
/cc @jmdobry, @tswast, @ace-n

### DIFF
--- a/samples/queries.js
+++ b/samples/queries.js
@@ -78,7 +78,6 @@ function queryStackOverflow(projectId) {
 // [END bigquery_simple_app_all]
 
 function syncQuery(sqlQuery, projectId) {
-  // [START bigquery_sync_query]
   // Imports the Google Cloud client library
   const BigQuery = require('@google-cloud/bigquery');
 
@@ -111,12 +110,10 @@ function syncQuery(sqlQuery, projectId) {
     .catch(err => {
       console.error('ERROR:', err);
     });
-  // [END bigquery_sync_query]
 }
 
 function asyncQuery(sqlQuery, projectId) {
-  // [START bigquery_async_query]
-  // [START bigquery_build_client]
+  // [START bigquery_query]
   // Imports the Google Cloud client library
   const BigQuery = require('@google-cloud/bigquery');
 
@@ -130,7 +127,6 @@ function asyncQuery(sqlQuery, projectId) {
   const bigquery = new BigQuery({
     projectId: projectId,
   });
-  // [END bigquery_build_client]
 
   // Query options list: https://cloud.google.com/bigquery/docs/reference/v2/jobs/query
   const options = {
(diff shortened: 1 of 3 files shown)


samples/queries.js
```diff
@@ -78,7 +78,6 @@ function queryStackOverflow(projectId) {
 // [END bigquery_simple_app_all]
 
 function syncQuery(sqlQuery, projectId) {
-  // [START bigquery_sync_query]
```
jmdobry: Don't delete this. This sample is displayed here: https://cloud.google.com/nodejs/

samples/queries.js
```diff
@@ -111,12 +110,10 @@ function syncQuery(sqlQuery, projectId) {
     .catch(err => {
       console.error('ERROR:', err);
     });
-  // [END bigquery_sync_query]
 }
 
 function asyncQuery(sqlQuery, projectId) {
-  // [START bigquery_async_query]
-  // [START bigquery_build_client]
+  // [START bigquery_query]
```
jmdobry: I assumed you'll be updating https://cloud.google.com/bigquery/docs/running-queries and https://cloud.google.com/bigquery/docs/reference/standard-sql/enabling-standard-sql with the new region tag?

samples/tables.js
```diff
@@ -566,7 +518,7 @@ function extractTableToGCS(
   filename,
   projectId
 ) {
-  // [START bigquery_extract_gcs]
+  // [START bigquery_extract_table]
```
jmdobry: I assume you'll be updating https://cloud.google.com/bigquery/docs/exporting-data to use the new region tag?

samples/tables.js
```diff
@@ -610,11 +562,11 @@ function extractTableToGCS(
     .catch(err => {
       console.error('ERROR:', err);
     });
-  // [END bigquery_extract_gcs]
+  // [END bigquery_extract_table]
 }
 
 function insertRowsAsStream(datasetId, tableId, rows, projectId) {
-  // [START bigquery_insert_stream]
+  // [START bigquery_table_insert_rows]
```
jmdobry: I assume you'll be updating https://cloud.google.com/bigquery/docs/streaming-data-into-bigquery to use the new region tag?

samples/queries.js
alixhami: Updated the sample include code on devsite to specify a branch tag so the sample render won't break. After this PR is merged, I will update devsite to use bigquery_query instead of bigquery_sync_query.

samples/queries.js
alixhami: Yes

samples/tables.js
alixhami: Yes

samples/tables.js
alixhami: Yes

PR #119, opened 2018-06-12
Added support for the NUMERIC values.

aprgoog:
This change adds support for the newly added NUMERIC data type to the nodejs BigQuery library.


JustinBeckwith:
Good question for @googleapis/node-team :)

ofrobots:
What's the forward compatibility story with the core [BigInt](https://developers.google.com/web/updates/2018/05/bigint) language feature that we just shipped with Node 10?

tswast:
@ofrobots Do BigInts support fractional values?

ofrobots:
It does not. I've removed the block. Will take a deeper look later.

ofrobots:
For posterity here is a link to the BigQuery docs on the NUMERIC data type: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#numeric-type

stephenplusplus:
LGTM, but we should probably have a system test for this new type.

aprgoog:
@stephenplusplus Added a system test, please take another look.

aprgoog:
Can somebody merge the changes? I do not have write access to the repo.

nikhilag:
I upgraded the bigquery node js library aftera long time and my app broke because of numeric values being returned as Big() values. Is it possible for me to have the existing behaviour where I got the numeric result back as string? I am storing the query result in firestore which won't support Big() and converting the results manually to string seems tedious. Is there any query option one can use to disable this new behaviour?

PR #146, opened 2018-07-12
chore(package): update to the latest @google-cloud/common

JustinBeckwith:


test/job.js
```diff
@@ -94,7 +76,7 @@ describe('BigQuery/Job', function() {
     });
 
     it('should promisify all the things', function() {
-      assert(promisified);
+      assert(job.getQueryResults.promisified_);
```
stephenplusplus: This is a hard one to follow. It depends on pseudo-private functionality in a different library, and we're only checking one method. Can we do a stub on `common.util.promisifyAll`?

JustinBeckwith:
@stephenplusplus was able to just get rid of this all together 😄

stephenplusplus (approved):

PR #175, opened 2018-08-24
chore: make ci happy

### LINKED ISSUE #174: Fix the nightly tests
The samples tests and system tests are both currently red:
https://circleci.com/gh/googleapis/workflows/nodejs-bigquery

callmehiphop:
Fixes #174
Closes #173 

- [x] Tests and linter pass

The original issue being addressed (#174) was remedied by simply upgrading the common package. However additional bugs were identified after that were introduced via #85. A second commit to fix said bug was created.

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -22,6 +22,7 @@ import {promisifyAll} from '@google-cloud/promisify';
 import * as extend from 'extend';
 import * as is from 'is';
 import {Table} from './table';
+import * as request from 'request';
 
 /**
  * Interact with your BigQuery dataset. Create a Dataset instance with
@@ -205,6 +206,7 @@ class Dataset extends ServiceObject {
       baseUrl: '/datasets',
       id: id,
       methods: methods,
+      requestModule: request,
       createMethod: (id, options, callback) => {
         if (is.fn(options)) {
           callback = options;
--- a/src/table.ts
+++ b/src/table.ts
@@ -27,6 +27,7 @@ const format = require('string-format-obj');
 import * as fs from 'fs';
 import * as is from 'is';
 import * as path from 'path';
+import * as request from 'request';
 const streamEvents = require('stream-events');
 import * as uuid from 'uuid';
 
@@ -242,6 +243,7 @@ class Table extends ServiceObject {
       id: id,
       createMethod: dataset.createTable.bind(dataset),
       methods: methods,
+      requestModule: request,
     });
 
     if (options && options.location) {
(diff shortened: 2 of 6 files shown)


src/dataset.ts
```diff
@@ -205,6 +206,7 @@ class Dataset extends ServiceObject {
       baseUrl: '/datasets',
       id: id,
       methods: methods,
+      requestModule: request,
```
stephenplusplus: We should make the common module inherit the parent's `requestModule`. @fhinkel wdyt?

callmehiphop: Makes sense to me!

src/table.ts
```diff
@@ -1114,53 +1116,8 @@ class Table extends ServiceObject {
...
-   *   .on('job', (job) => {
-   *     // `job` is a Job object that can be used to check the status of the
-   *     // request.
-   *   })
-   *   .on('complete', (job) => {
-   *     // The job has completed successfully.
-   *   });
    */
-  createWriteStream(metadata) {
+  createWriteStream_(metadata) {
```
stephenplusplus: Why the separation?

callmehiphop: We call `createWriteStream` within `createLoadJob`, however with the addition of https://github.com/googleapis/nodejs-bigquery/pull/85/files#diff-6494577af08481767e965304488d997cR1253, we introduced a race condition that could fire the jobs complete event off before the job was actually returned to the user.

We want to wait to start polling the job until the user starts, so I think this would be the best way to go in order to preserve the changes from #85.

callmehiphop:
@JustinBeckwith is there something special I need to do to get kokoro passing?

JustinBeckwith (approved):
LGTM, but let @stephenplusplus give it a 👍 too :)

src/dataset.ts
fhinkel: Instead of specifying it in the constructor? Not sure I see a huge advantage, as you need to specify it either way.

stephenplusplus: Sorry, my terminology was a bit off. `Dataset` is a `ServiceObject`, whose parent is `BigQuery`, a `Service`. Since we set `requestModule` through BigQuery, it would be easier if the ServiceObject children would inherit it to save us from the multiple imports and redundancy.

stephenplusplus: Opened an issue on common: https://github.com/googleapis/nodejs-common/issues/222

stephenplusplus (approved):

PR #177, opened 2018-08-27
fix: (tests) Use a filter to locate datasets used in tests.

### LINKED ISSUE #176: System tests might be deleting all datasets inside our test project
https://github.com/googleapis/nodejs-bigquery/blob/960d84e3e320685d3766508e6d8a3988c1b6397e/system-test/bigquery.ts#L1672-L1680

There's no reference to `prefix` option to list datasets:
https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/list

And we're suspecting it is causing a test dataset in nodejs-dlp to be deleted, causing its nightly to fail: 
https://github.com/googleapis/nodejs-dlp/issues/52#issuecomment-415837211

stephenplusplus:
Fixes #176

### DIFF
--- a/system-test/bigquery.ts
+++ b/system-test/bigquery.ts
@@ -32,7 +32,7 @@ const bigquery = new BigQuery();
 const storage = require('@google-cloud/storage')();
 
 describe('BigQuery', function() {
-  const GCLOUD_TESTS_PREFIX = 'gcloud_test_';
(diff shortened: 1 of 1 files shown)


kinwa91:
Please continue to ignore kokoro failures. It's not properly configured here yet.

PR #188, opened 2018-09-08
Use arrow functions

JustinBeckwith:


JustinBeckwith:
@kinwa91 the tests are failing here because it looks like we require an argument in the `Storage` constructor 😓

PR #193, opened 2018-09-17
Use async/await in samples

JustinBeckwith:
Mostly doing this to start a conversation :) What do y'all think of using async/await in our samples?  I figure, we tell people to start with the latest LTS version of nodejs, which is 8 at this stage 🤷‍♂️  Node.js 6 is 6 months from death, so I don't feel too bad doing this just in the samples. 

Thoughts?

@googleapis/node-team @fhinkel @ace-n

### DIFF
--- a/samples/datasets.js
+++ b/samples/datasets.js
@@ -15,7 +15,7 @@
 
 'use strict';
 
-function createDataset(datasetId, projectId) {
+async function createDataset(datasetId, projectId) {
   // [START bigquery_create_dataset]
   // Imports the Google Cloud client library
   const BigQuery = require('@google-cloud/bigquery');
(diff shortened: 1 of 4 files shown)


ace-n:
FWIW the GCF samples include both 6 and 8 (where applicable).

Do we have any relevant data here? (Ping us internally if it's non-public.)

ofrobots:
At this point it makes sense for the samples to assume async/await support. Node 6 will be EOL in April. It is declining in usage (see https://nodejs.org/metrics) and already Node 8 seems be getting 3x the downloads as Node 6.

fhinkel:
Yeah, I think we can start with `async/await` samples.

JustinBeckwith:
@fhinkel what are your thoughts on the way I eliminated the error handling?  We used to console.log the errors, now I am just letting the async method throw.  Is this the pattern we want to roll with for our samples?

JustinBeckwith:
So anyone willing to approve it then? 🤣

JustinBeckwith:
🤷‍♂️ I'll take it

PR #227, opened 2018-10-25
refactor(typescript): convert index to es module

JustinBeckwith:
BREAKING CHANGE: The import style of this library has been changed to be es module compatible.

### Old Code
```js
const BigQuery = require('@google-cloud/bigquery');
```

### New Code
```js
const {BigQuery} = require('@google-cloud/bigquery');
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "pretest": "npm run compile"
   },
   "dependencies": {
-    "@google-cloud/common": "^0.25.0",
+    "@google-cloud/common": "^0.26.0",
     "@google-cloud/paginator": "^0.1.0",
     "@google-cloud/promisify": "^0.3.0",
     "arrify": "^1.0.0",
@@ -87,6 +87,7 @@
     "@google-cloud/nodejs-repo-tools": "^2.2.3",
     "@google-cloud/storage": "^2.0.0",
     "@types/mocha": "^5.2.5",
+    "@types/sinon": "^5.0.5",
     "async": "^2.6.0",
     "codecov": "^3.0.0",
     "eslint": "^5.0.0",
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "benchmark": "time node benchmark/bench.js benchmark/queries.json",
     "docs": "jsdoc -c .jsdoc.js",
     "lint": "eslint samples/",
-    "prettier": "prettier --write samples/*.js samples/*/*.js",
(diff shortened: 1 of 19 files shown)


test/index.ts
```diff
@@ -410,13 +420,13 @@ describe('BigQuery', () => {
       day: 1,
     };
 
-    it('should expose static and instance constructors', () => {
+    it.skip('should expose static and instance constructors', () => {
```
JustinBeckwith: @stephenplusplus @callmehiphop so this is broke.  Instead of `bigquery.date` referring to a type, a static method, and an instance method - I'm now just exporting `BigQueryDate` as a type (same for the other similar objects), and keeping the instance and static methods. 

Thoughts on the approach?

src/table.ts
```diff
@@ -18,7 +18,7 @@
 
 import * as arrify from 'arrify';
 import * as Big from 'big.js';
-import {ServiceObject, util, SetMetadataResponse, ResponseCallback} from '@google-cloud/common';
+import * as common from '@google-cloud/common';
```
kinwa91: Any particular reason for this?

test/index.ts
kinwa91: gentle ping @stephenplusplus @callmehiphop on your take here :)

kinwa91: Also, if this is a change in API, please document what was changed and how users should update their code

src/table.ts
JustinBeckwith: Haha, yeah.  The way we appear to be using proxyquire in the tests relies on stubbing `util` specifically.  If I separated it out, things stopped working.  We can revisit this, but with the size of the change I was really just looking to get things healthy again.

JustinBeckwith:
Tip to reviewers @kinwa91 @stephenplusplus - you probably want to add `?w=1` in the url to filter out all the whitespace junk and noise:
https://github.com/googleapis/nodejs-bigquery/pull/227/files?w=1

package.json
```diff
@@ -57,7 +57,6 @@
     "benchmark": "time node benchmark/bench.js benchmark/queries.json",
     "docs": "jsdoc -c .jsdoc.js",
     "lint": "eslint samples/",
```
kinwa91: Also run `gts check` here?

`"lint": "gts check && eslint samples/"`

JustinBeckwith: Can't yet :/

PR #270, opened 2018-11-20
Changing import of Big from big.js so it doesn't use default

### LINKED ISSUE #269: TypeError in production version using query params

#### Environment details

  - OS: Windows
  - Node.js version: 10.13.0
  - npm version: 6.4.1
  - `@google-cloud/bigquery` version: 2.0.1

#### Steps to reproduce

```js
const dataset = BigQueryInstance.dataset("dataset");
await dataset.query({ query: "SELECT ... WHERE id =?", params: [ id ] });
```

Resultant error when attempting to execute

```
Error 
TypeError: Right-hand side of 'instanceof' is not an object
    at Function.getType_ (...\node_modules\@google-cloud\bigquery\build\src\index.js:462:1)
```

The production version, received from npm install, on the function "static getType_(value) {" has been compiled to

```
462:        else if (value instanceof big_js_1.default) {

this is "undefined"
```

and quite different to the

```
616:    } else if (value instanceof Big) {
```

of the source code.

The Big object is defined in the big_js_1 object, I am unsure why the compiler has done this?

negatron99:
Fixes #269 (it's a good idea to open an issue first for discussion)

- [X] Tests and linter pass
- [n/a] Code coverage does not decrease (if any source code was changed)
- [n/a] Appropriate docs were updated (if necessary)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,7 +18,7 @@ import * as common from '@google-cloud/common';
 import {paginator} from '@google-cloud/paginator';
 import {promisifyAll} from '@google-cloud/promisify';
 import * as arrify from 'arrify';
-import Big from 'big.js';
+import {Big} from 'big.js';
 import * as extend from 'extend';
 
 const format = require('string-format-obj');


negatron99:
I signed it!

PR #280, opened 2018-12-03
docs: Improve timestamp documentation.

crwilcox:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -585,7 +585,7 @@ export class BigQuery extends common.Service {
    * zone or convention such as Daylight Savings Time.
    *
    * @method BigQuery#timestamp
-   * @param {date} value The time.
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -585,7 +585,7 @@ export class BigQuery extends common.Service {
    * zone or convention such as Daylight Savings Time.
    *
    * @method BigQuery#timestamp
-   * @param {date} value The time.
+   * @param {date} value The time represented by a `Date` or `string`
```
JustinBeckwith: Not to nitpick, but shouldn't we update the jsdoc type too?

crwilcox: sure!

PR #291, opened 2018-12-06
fix(docs): change source location to ./build for ts project

jkwlui:


synth.py
```diff
@@ -4,5 +4,5 @@
 
 logging.basicConfig(level=logging.DEBUG)
 common_templates = gcp.CommonTemplates()
-templates = common_templates.node_library()
+templates = common_templates.node_library(source_location='build')
```
JustinBeckwith: Not build/src?

PR #295, opened 2018-12-07
fix: fix typescript compilation

### LINKED ISSUE #287: @google-cloud/storage should be defined as a normal dependency
Version 2.0.2 has an import of the "@google-cloud/storage" package inside `table.ts`. However it's defined as a `devDependency` instead of a `dependency` in `package.json`. This forces me to introduce this dependency to my own project or it will cause a typescript compile error.

See https://github.com/googleapis/nodejs-bigquery/blob/master/src/table.ts#L36
It was added with this PR https://github.com/googleapis/nodejs-bigquery/pull/246

Thanks!

JustinBeckwith:
Fixes #287.  There are a few places where we accept `File` type objects, which is defined in `google-cloud/storage`.  The package is only needed for types, so it's installed as a `devDependency`.  However - this means the types for storage aren't shipped with the module, which causes compilation to fail for consumers of the package.  This change dumps the `File` type for now in favor of `any`, and adds an install test to verify basic compilation works after an `npm pack`.


JustinBeckwith:
@googleapis/node-team this is an interesting problem, and I'm interested in y'alls opinions :)

ofrobots:
I see this problem with TypeScript code once per fortnight. The real solution here would be a way to tell the package installer this info: "If you install this package as a devDependency of a top-level install, you also need to install these other devDependencies of mine".

IOW, we need a `transitiveDevDependencies` section in `package.json` that somehow npm could respect. It might be worth engaging with TypeScript and npm folks on this.

PR #311, opened 2018-12-19
Release @google-cloud/bigquery v2.0.4

kinwa91:
As with any changes in types, I'm a little held back whether it should qualify as a patch vs a minor release. Please let me know if a patch release here doesn't make sense.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,29 @@
 
 [1]: https://www.npmjs.com/package/@google-cloud/bigquery?activeTab=versions
 
+## v2.0.4
+
+12-19-2018 14:35 PST
+
+### Implementation Changes
+- fix(ts): explicit usage of Duplex ([#300](https://github.com/googleapis/nodejs-bigquery/pull/300))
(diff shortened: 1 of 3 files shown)


kinwa91:
This test has been flaky:
```
  1) BigQuery
       "before all" hook for "should be able to use the d.ts":
     Error: Invalid value at 'dataset' (Map), Cannot bind a list to map for field 'labels'.
      at Util.parseHttpRespBody (node_modules/@google-cloud/common/build/src/util.js:172:38)
      at Util.handleResp (node_modules/@google-cloud/common/build/src/util.js:116:117)
      at retryRequest (node_modules/@google-cloud/common/build/src/util.js:403:22)
      at onResponse (node_modules/retry-request/index.js:200:7)
      at /tmpfs/src/github/nodejs-bigquery/node_modules/teeny-request/src/index.ts:237:19
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)

```

JustinBeckwith:
While the types are nascent, don't think it's reasonable to do semver major releases.  Patch is fine here.

PR #317, opened 2018-12-21
fix(test): skip flaky invalid etag test

kinwa91:


JustinBeckwith:
@kinwa91 please make sure we have an issue in the tracker to investigate and turn this back on :)

PR #319, opened 2018-12-27
fix: remove Job.setMetadata method

jkwlui:
BigQuery/Jobs API doesn't (no longer?) support PATCH method on jobs.

I'm not sure if it used to, @stephenplusplus, but the REST [reference](https://cloud.google.com/bigquery/docs/reference/rest/v2/) doesn't show a PATCH method.

When attempting to call Job.setMetadata, the promise rejects citing "Unable to parse JSON" because the server returned a 404 HTML response saying the method is not found.

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -221,45 +221,6 @@ class Job extends Operation {
           qs: {location},
         },
       },
-
-      /**
-       * Set the metadata for this job. This can be useful for updating job
-       * labels.
-       *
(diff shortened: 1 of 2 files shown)


stephenplusplus (approved):
This is kind of weird. I would guess we enabled `setMetadata` by mistake for the Jobs API.

crwilcox (approved):

PR #323, opened 2019-01-08
fix: correctly iterate query results within stream

### LINKED ISSUE #303: Is createQueryStream removed or is it just not working in v2
I was using `createQueryStream` in the 1.3 version. After updating to v2 the call does not work correctly anymore. It seems to start but then something goes wrong (see pic). The query string is passed in and is valid as it can be logged before the call and v1.3 works.

The v2 documentation does not mention `createQueryStream` anymore. Does that mean it was removed? The removal is not marked as a breaking change in the changelog.

image

#### Environment details

  - OS: Running in GAE
  - Node.js version: 10
  - npm version:
  - `@google-cloud/bigquery` version: 2

#### Steps to reproduce

  1. Use `createQueryStream` in v1.3
  2. Update to v2. Call seems to start but soon causes errors.

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

stephenplusplus:
Fixes #303 

In a stream, we-- y'know what, @oliverwoodings explained it best: https://github.com/googleapis/nodejs-bigquery/issues/303#issuecomment-452288078

> - My code calls `bigquery.createQueryStream()` with a query. Note that this function has been wrapped by @google-cloud/paginator
> - This calls `queryAsStream_` (https://github.com/googleapis/nodejs-bigquery/blob/master/src/index.ts#L1515), which ends up calling `createQueryJob`
> - After we have a job, we call `job.getQueryResults` (https://github.com/googleapis/nodejs-bigquery/blob/master/src/index.ts#L1505).
> - After getting the job status, the job i'm running hasn't completed, so we call back with no rows and no page token (https://github.com/googleapis/nodejs-bigquery/blob/master/src/job.ts#L418-L428)
> - The `callback` given to `getQueryResults` that is called at this point is actually the results handler from @google-cloud/paginator (https://github.com/googleapis/nodejs-paginator/blob/master/src/index.ts#L315)
> - Because there is a `nextQuery` given to `onResultSet` of the paginator, and there are still rows to send, the paginator executes the request again (https://github.com/googleapis/nodejs-paginator/blob/master/src/index.ts#L332-L335)
> - `makeRequest` in the paginator calls `originalMethod`, which in this case is actually `BigQuery.queryAsStream_`, which means it tries to create an entire new query job, instead of just polling for more results?!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -51,6 +51,7 @@ export interface SimpleQueryRowsCallback {
 export interface Query {
   dryRun?: boolean;
   location?: string;
+  job?: Job;
   jobId?: string;
   jobPrefix?: string;
   // tslint:disable-next-line no-any
@@ -1489,7 +1490,7 @@ export class BigQuery extends common.Service {
       QueryRowsCallback,
       cb?: SimpleQueryRowsCallback|QueryRowsCallback):
       void|Promise<SimpleQueryRowsResponse>|Promise<QueryRowsResponse> {
-    const options =
+    let options =
         typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
     const callback =
         typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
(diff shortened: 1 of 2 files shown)


stephenplusplus:
@oliverwoodings it's your patience going through all of that that is most impressive :D

PR #331, opened 2019-01-17
fix(types): add missing parameters (selectedFields, startIndex) in table.getRows() options

### LINKED ISSUE #330: table.getRows() has missing options in its types
#### Environment details

#### Steps to reproduce

  1. Try use `table.getRows(options)`
  2. Can't pass in `selectedFields` or `startIndex`, as is possible with the direct HTTP endpoint.

I've confirmed that passing those fields into the function does work as expected (at least in my small use case).

I see that behind the scenes its calling the BigQuery HTTP API directly, so I assume these fields are possible, should work as expected and are just missing the the current types.

lostpebble:
Fixes #330

- [x] Tests and linter pass
- [] Appropriate docs were updated (if necessary)

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -82,6 +82,8 @@ export interface TableRowField {
 export type TableRowValue = string|TableRow;
 
 export interface GetRowsOptions {
+  startIndex?: number;
+  selectedFields?: string;
   autoPaginate?: boolean;
   maxApiCalls?: number;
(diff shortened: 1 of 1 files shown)


lostpebble:
I signed it!

lostpebble:
No worries! I was happy to make the relevant changes to the docs too, but not sure exactly where to do that.

PR #333, opened 2019-01-22
fix(types): Allows views to be configured using an object or a string

### LINKED ISSUE #332: Types (e.g. TableMetadata) disallow valid API usage
#### Environment details

  - OS: OSX 10.13
  - Node.js version: v11.7.0
  - npm version: 6.5.0

#### Steps to reproduce

Write this code:

```ts
dataset.createTable(
      'name'
      {
        view: {
          query: 'any SQL string',
          useLegacySql: false,
        },
      } as TableMetadata,
    )
```

The Typescript compiler will complain and say that `friendlyName` is required, and `view` must be a string. [The types in the API documentation](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#resource) say that friendlyName is optional, and view is an object with some properties.

I discovered this when updating from v1.3 of the library to v2. I'd expect my existing, working library calls to pass type checking, since everything still works if I disable type checking and run the code.

Tyler-Murphy:
Previously, only a string containing the view definition query could be provided. This allows an object, as described in the API documentation: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#resource.

Fixes #332

- [x] Tests and linter pass
- [ ] ~~Appropriate docs were updated (if necessary)~~

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -161,7 +161,12 @@ export interface TableMetadata {
   friendlyName: string;
   schema?: string|TableField[];
   partitioning?: string;
-  view?: string;
+  view?: string|ViewDefinition;
+}
+
+export interface ViewDefinition {
+  query: string;
+  useLegacySql?: boolean;
 }
 
 export interface FormattedMetadata {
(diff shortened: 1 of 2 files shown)


Tyler-Murphy:
I signed it!

PR #339, opened 2019-01-29
build: output benchmark data in csv format

JustinBeckwith:


tswast:
@JustinBeckwith  Ooh. Are these going to be dashboarded anywhere?

JustinBeckwith:
So this was just for my purposes.  I was trying to do some trending by hand, and having it pipe-able to CSV was making things easier.  After this though...  I've got this idea in my head for something like codecov, but specifically for benchmarks.  You could run a centralized service that has the GitHub API integration, and a common metadata format.  Then, you write clients for creating the benchmarks in each language.

Then I realized that was a lot of work and played nintendo instead 🤷‍♂️

PR #345, opened 2019-02-05
test: skip installation test if GOOGLE_CLOUD_TESTS_IN_VPCSC is passed

jkwlui:
Split out install-test as a separate npm run-script instead of bundling
with system-test.

This helps unblock VPCSC testing since it can't fetch external
dependencies with npm install when installation test does npm pack


callmehiphop (approved):

JustinBeckwith (changes requested):
Instead of this - could we do the thing you did in storage where we looked for an env var, and did an explicit skip?  I don't want the kokoro configs straying if we can at all avoid it.

kinwa91:
I changed it to use the environment variable to determine whether or not to skip install-test, but I still kept it split out

JustinBeckwith (approved):

PR #363, opened 2019-02-28
chore: make test prefix unique per run

callmehiphop:
We're seeing a bunch of seemingly random 404 errors in our CI system test job. I believe this is because we now have to 2 system-test jobs and our test prefix is not unique. I was going to use `uuid` but doing so puts us over the character limit.

### DIFF
--- a/system-test/bigquery.ts
+++ b/system-test/bigquery.ts
@@ -1523,8 +1523,9 @@ describe('BigQuery', () => {
   });
 
   function generateName(resourceType: string) {
-    return `${GCLOUD_TESTS_PREFIX}_${resourceType}_${uuid.v1()}`.replace(
-        /-/g, '_').substr(0, 63);
+    return `${GCLOUD_TESTS_PREFIX}_${resourceType}_${uuid.v1()}`
+        .replace(/-/g, '_')
+        .substr(0, 63);
   }
 
   async function deleteBuckets() {
--- a/system-test/bigquery.ts
+++ b/system-test/bigquery.ts
@@ -28,7 +28,7 @@ const bigquery = new BigQuery();
 const storage = new Storage();
 
 describe('BigQuery', () => {
-  const GCLOUD_TESTS_PREFIX = 'nodejs_bq_test';
+  const GCLOUD_TESTS_PREFIX = `nodejs_bq_test${Date.now()}`;
 
   const dataset = bigquery.dataset(generateName('dataset'));
   const table = dataset.table(generateName('table'));
@@ -1524,7 +1524,7 @@ describe('BigQuery', () => {
 
   function generateName(resourceType: string) {
     return `${GCLOUD_TESTS_PREFIX}_${resourceType}_${uuid.v1()}`.replace(
-        /-/g, '_');
+        /-/g, '_').substr(0, 63);
   }
 
   async function deleteBuckets() {


system-test/bigquery.ts
```diff
@@ -28,7 +28,7 @@ const bigquery = new BigQuery();
 const storage = new Storage();
 
 describe('BigQuery', () => {
-  const GCLOUD_TESTS_PREFIX = 'nodejs_bq_test';
+  const GCLOUD_TESTS_PREFIX = `bq${Date.now()}`;
```
stephenplusplus: Can we keep some `nodejs_` reference in there. This helped us not collide with other projects, as well as just being friendlier when we need to trace where a straggler came from.

callmehiphop: Fair enough, we'll probably have to change the way we generate the resource names via `generateName` though because we're pretty much at the limit just by adding the timestamp.

JustinBeckwith: In other libs we're doing a lot of `gcloud-test-` or something to that effect.

stephenplusplus: We were getting in trouble with the generic prefix when our before/after hooks filtered previously-created resources. It would actually delete the other projects'! I believe we added in some more logic about how old a resource must be before we consider it dead, but I'm not sure which repos we have done that in. At a minimum, a prefix unique to the language of the project eliminates most chances for collisions.

JustinBeckwith: Oh bother - I meant something like `gcloud-test-${uuid.v4().split('-')[0]}`

stephenplusplus: The collision potential is still there, because our test hooks filter on the first part of the prefix (the preprefixfix, in technical terms), like `gcloud-test-`, so that we can get a list of all resources created by the project without knowing the full name.

callmehiphop: So I've removed including the resource type in the name and it would appear we're staying under the limit.

stephenplusplus: That's good except for test failure, cause it can be super confusing figuring out what "Cannot delete ___random_name__" means :(

callmehiphop: Where would that message come from though? In BQ we only delete Buckets and Datasets, I'm not particularly familiar with what error messages would be for either, but doing a quick test Storage doesn't seem to include the bucket name for not found or permission errors.

stephenplusplus: Hmm, I'm not sure specifically the errors for these libs, but I recall it being very important with Compute, because many different resource types are deleted within the same after hook, and it says "blah cannot be deleted". Can we just clip more off of the uuid? We can play with the prefix, too, e.g. `bqnode_{type}_{uuid}`

callmehiphop: Ok, made a new commit. The hard limit for Buckets is 63 so I just added a `.substr(0, 63)` to the end of the generated name.

JustinBeckwith: Another easy idea - for the randomized part, use `Math.random().toString(36).slice(2)`  👅

stephenplusplus:
Ok! I figured out what is going on. We shouldn't change the prefix that we filter on. We use that to delete accidentally-created, stale datasets from previous test runs. Instead, we should only delete ones that are > 24 hours old, for example. That way, we'll avoid parallel CI-run collisions.

PR #365, opened 2019-03-01
fix(types): move JobLoadMetadata writeDisposition

arjenvanderende:
The `writeDisposition` field for interface `JobLoadMetadata` appears to have been defined incorrectly; it does not belong under `timePartitioning`.

Typescript fails to compile if I define it under the root, but works as expected if I cast the options to `any` like so:

```javascript
new BigQuery({ projectId: "my-project-id" })
    .dataset("my_dataset")
    .table("my_table")
    .load("my_file.csv", {
        format: "csv",
        writeDisposition: "WRITE_TRUNCATE",
    } as any);
```

See:
https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load

- [ ] Tests and linter pass

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -119,8 +119,8 @@ export interface JobLoadMetadata {
     field?: string;
     requirePartitionFilter?: boolean;
     type?: string;
-    writeDisposition?: string;
   };
+  writeDisposition?: string;
 }
 
 export interface CreateExtractJobOptions {


src/table.ts
```diff
@@ -119,8 +119,8 @@ export interface JobLoadMetadata {
     field?: string;
     requirePartitionFilter?: boolean;
     type?: string;
-    writeDisposition?: string;
   };
+  writeDisposition?: 'WRITE_TRUNCATE'|'WRITE_APPEND'|'WRITE_EMPTY';
```
stephenplusplus: Thank you for this. Would you mind taking out the specific string values, however? The upstream is the official resource for acceptable values, and we don't want to duplicate from this layer in the event it is changed upstream.

arjenvanderende: No problem. I think it improve typing, but I understand your reasoning for wanting to keep it simple.

Keep in mind, I copied the definitions from `CopyTableMetadata` which also includes specific string values. 😉 
See: https://github.com/googleapis/nodejs-bigquery/blob/master/src/table.ts#L151-L157

I changed the type back to an optional `string` and amended/rebased the commit.

stephenplusplus: Oh, gotcha. I didn't realize it was there, too! Thank you for all of the help!

PR #368, opened 2019-03-06
refactor(samples): split samples into their own files

### LINKED ISSUE #367: update samples to follow canonical
Update samples to follow new standards, including separating methods and test into individual files.

steffnay:
Fixes #367

- [ ] Tests and linter pass


samples/bigquery_create_dataset.js
```diff
@@ -0,0 +1,40 @@
...
+// [START bigquery_create_dataset]
+async function bigquery_create_dataset(
+  PROJECT_ID = 'YOUR_PROJECT_ID',
+  DATASET_ID = 'YOUR_DATASET_ID'
+) {
+  // Imports the Google Cloud client library
+  const {BigQuery} = require('@google-cloud/bigquery');
+
+  // Creates a client
+  const bigquery = new BigQuery({PROJECT_ID});
```
JustinBeckwith: There's no real need to include the `{PROJECT_ID}` part here, we can just drop that

samples/bigquery_create_dataset.js
```diff
@@ -0,0 +1,40 @@
...
+
+  // Creates a new dataset
+  await bigquery
```
JustinBeckwith: In an async function, you don't have to do it this way!  Instead:

```js
const [dataset] = await bigquery.createDataset(DATASET_ID);
console.log(`Dataset ${dataset.id} created.`);
```

No need to do any of the error handling stuff there either. Just let it throw!

samples/bigquery_create_dataset.js
JustinBeckwith: (same feedback for samples below)

samples/bigquery_create_dataset.js
JustinBeckwith: (same feedback for samples below)

samples/test/datasets.test.js
```diff
@@ -1,63 +0,0 @@
-/**
```
JustinBeckwith: I don't know if we talked about this in the rubric or not - but I'm actually fine with keeping the tests together.  No  need to separate all that out into individual files IMO.

samples/bigquery_create_dataset.js
steffnay: I'm curious about this because I had some struggles with it while friction logging the samples when they were formatted that way. For example, when I called createDataset with the same DATASET_ID twice in a row, I got an unclear error message within an unhandled promise rejection and I had to manually change the sample to figure out that the dataset had already existed. Is there another way to handle that scenario?

JustinBeckwith: Yes!  This line at the bottom:

```js
bigquery_list_datasets(...process.argv.slice(2));
```

Should be ...

```js
bigquery_list_datasets(...process.argv.slice(2)).catch(console.error);
```

For folks who pick this up and just run it, that should take care of things.  For others that are copy-pasting the snippets - your error handling management is something we have to balance.  While doing the try/catch inline makes it easier for people to see the error for a given snippet, it's *never* code you would want to write in production.  That would mask errors, and make the service act unpredictably.  Instead - you would just let it throw, like we're doing. 

If this doesn't make sense - let me know!

steffnay: Makes sense! Thank you!

samples/bigquery_create_dataset.js
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+// [START bigquery_create_dataset]
+async function bigquery_create_dataset(DATASET_ID = 'YOUR_DATASET_ID') {
+  // Imports the Google Cloud client library
```
tswast: Nit: In general we use imperatives as in, "you the developer should do this", as opposed to "this is what the code does."

This comment would become: "Import the Google Cloud client library"

samples/bigquery_delete_dataset.js
```diff
@@ -0,0 +1,35 @@
...
+  const {BigQuery} = require('@google-cloud/bigquery');
+
+  // Creates a client
+  const bigquery = new BigQuery();
+
+  // Creates a reference to the existing dataset
+  const dataset = bigquery.dataset(DATASET_ID);
+
+  // Deletes the dataset
+  await dataset.delete();
```
tswast: Just as with the Python sample, it'd be nice to show how to delete all the tables a dataset contains as part of the same request. (Python has delete_contents=True, but hopefully there's something similar in the Node API).

PR #377, opened 2019-03-11
testing: remove nextQuery assertion

### LINKED ISSUE #375: The system tests, they are broken.

callmehiphop:
Fixes #375 

Since we refactored our strategy for dealing with lingering resources, sometimes `nextQuery` will be populated with a page of resources from a different test run. 

- [x] Tests and linter pass

### DIFF
--- a/system-test/bigquery.ts
+++ b/system-test/bigquery.ts
@@ -176,7 +176,6 @@ describe('BigQuery', () => {
       filter: `labels.${GCLOUD_TESTS_PREFIX}`,
     });
     assert(datasets[0] instanceof Dataset);
-    assert.strictEqual(nextQuery, null);
     assert(apiResponse);
   });
 


system-test/bigquery.ts
```diff
@@ -176,7 +176,6 @@ describe('BigQuery', () => {
       filter: `labels.${GCLOUD_TESTS_PREFIX}`,
     });
     assert(datasets[0] instanceof Dataset);
-    assert.strictEqual(nextQuery, null);
```
bcoe: seems reasonable to me; in a perfect world, I might try to have a `beforeEach` that ensures that there's nothing populated in `nextQuery` before running the test. Is there a tracking issue for this bug? (state between tests can really start to be a pain).

callmehiphop: I don't think there is, @JustinBeckwith might know though?

PR #385, opened 2019-03-15
fix(ts): provide complete and correct types

### LINKED ISSUE #332: Types (e.g. TableMetadata) disallow valid API usage

callmehiphop:
Fixes #332

This basically rewrites all the types. Pretty much every request/response interface was missing fields and every callback/response returned `request.Response` which we don't return anywhere.
~In an effort to **not** make breaking changes, I've kept all type names and duplicate types in tact.~

- [x] Tests and linter pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "presystem-test": "npm run compile",
     "check": "gts check",
     "clean": "gts clean",
-    "compile": "tsc -p .",
+    "compile": "tsc -p . && cp src/types.d.ts build/src/",
     "fix": "gts fix && eslint --fix '**/*.js'",
     "predocs": "npm run compile",
     "prepare": "npm run compile",
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -15,15 +15,16 @@
  */
 
 import {DecorateRequestOptions, DeleteCallback, ServiceObject} from '@google-cloud/common';
-import {paginator} from '@google-cloud/paginator';
+import {paginator, ResourceStream} from '@google-cloud/paginator';
 import {promisifyAll} from '@google-cloud/promisify';
 import * as extend from 'extend';
 import * as r from 'request';
 import {Duplex, Readable} from 'stream';
 import {teenyRequest} from 'teeny-request';
 
-import {BigQuery, DatasetCallback, Query, QueryRowsResponse, SimpleQueryRowsCallback} from '.';
+import {BigQuery, DatasetCallback, PagedCallback, PagedRequest, PagedResponse, Query, QueryRowsResponse, ResourceCallback, SimpleQueryRowsCallback} from '.';
 import {JobCallback, JobResponse, Table, TableMetadata, TableOptions} from './table';
+import {bigquery} from './types';
 
 export interface DatasetDeleteOptions {
   force?: boolean;
@@ -33,25 +34,16 @@ export interface DataSetOptions {
   location?: string;
 }
 
-export interface CreateDatasetOptions {}
+export type CreateDatasetOptions = bigquery.IDataset;
 
-export interface GetTablesOptions {
-  autoPaginate?: boolean;
-  maxApiCalls?: number;
-  maxResults?: number;
-  pageToken?: string;
-}
+export type GetTablesOptions = PagedRequest<bigquery.IListTablesRequest>;
+export type GetTablesResponse =
+    PagedResponse<Table, GetTablesOptions, bigquery.IListTablesResponse>;
+export type GetTablesCallback =
+    PagedCallback<Table, GetTablesOptions, bigquery.IListTablesResponse>;
 
-export type GetTablesResponse = [Table[], r.Response];
-export interface GetTablesCallback {
-  (err: Error|null, tables?: Table[]|null, nextQuery?: {}|null,
-   apiResponse?: r.Response): void;
-}
-
-export type TableResponse = [Table, r.Response];
-export interface TableCallback {
-  (err: Error|null, table?: Table|null, apiResponse?: r.Response): void;
-}
+export type TableResponse = [Table, bigquery.ITable];
+export type TableCallback = ResourceCallback<Table, bigquery.ITable>;
 
 /**
  * Interact with your BigQuery dataset. Create a Dataset instance with
@@ -72,7 +64,7 @@ export interface TableCallback {
 class Dataset extends ServiceObject {
   bigQuery: BigQuery;
   location?: string;
-  getTablesStream: () => Readable;
+  getTablesStream: (options?: GetTablesOptions) => ResourceStream<Table>;
   constructor(bigQuery: BigQuery, id: string, options?: DataSetOptions) {
     const methods = {
       /**
@@ -298,7 +290,7 @@ class Dataset extends ServiceObject {
      *     this.end();
      *   });
      */
-    this.getTablesStream = paginator.streamify('getTables');
+    this.getTablesStream = paginator.streamify<Table>('getTables');
   }
 
   createQueryJob(options: string|Query): Promise<JobResponse>;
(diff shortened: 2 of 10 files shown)


JustinBeckwith:
Uh, is this a breaking change for TypeScript users?

package.json
```diff
@@ -37,7 +37,7 @@
     "presystem-test": "npm run compile",
     "check": "gts check",
     "clean": "gts clean",
-    "compile": "tsc -p .",
+    "compile": "tsc -p . && cp src/types.d.ts build/src/",
```
JustinBeckwith: To keep things straight in the books - what do you think of installing `googleapis` as a dev dependency, then having this step copy the `d.ts` out of there?  Similar to what we do with `pbts`?

callmehiphop:
@JustinBeckwith maybe? I'm not really sure how to classify changing interfaces that were flat out wrong.

package.json
callmehiphop: Oh, I would be cool with that - I didn't realize that there were any protos/declarations available in `googleapis`.

JustinBeckwith:
Hah, I hear you.  For now let's call it semver patch I guess?

callmehiphop:
@JustinBeckwith sounds good.. like I said in the overview my only concern would be about some fields being marked optional now. But I could probably just use a generic to mark them as non-optional until we're ready for a semver major.

package.json
callmehiphop: So I didn't find anything like what you were mentioning for BigQuery in googleapis. I'd also have some concerns about generating types via `pbjs` and `pbts` because I think it means we'd need to include protobufjs/long as a dependency, which doesn't make sense for BigQuery yet.

callmehiphop: Ohh, I thought you meant the googleapis *repo*, my bad.

I'm kind of less into this, it won't be as simple as a `cp` since that file isn't just types, but also a class that requires an additional dep to pull in external types and functions. Additionally some of the types have extra fields that are specific to `googleapis` (namely `auth`) which don't really apply here. 

The types I have here are just a copy/paste of those minus the `googleapis` specific fields and renamed to resemble the types in our grpc clients.

bcoe: I've always been advised against copying files out of `node_modules`, just because there's no guarantee it will be structured as you expect (as an example, perhaps there's two versions of a library, so the library you want is actually nested in a sub-folder -- in this case it doesn't see like that could happen, just always something I'm paranoid about.

If you do go this route, perhaps make a little build script that uses:

```
require.resolve('googleapis/build/srce/apis/bigquery/v2.d.ts')
```

src/dataset.ts
```diff
@@ -15,15 +15,16 @@
...
 import {promisifyAll} from '@google-cloud/promisify';
 import * as extend from 'extend';
 import * as r from 'request';
 import {Duplex, Readable} from 'stream';
 import {teenyRequest} from 'teeny-request';
 
-import {BigQuery, DatasetCallback, Query, QueryRowsResponse, SimpleQueryRowsCallback} from '.';
+import {BigQuery, DatasetCallback, PagedCallback, PagedRequest, PagedResponse, Query, QueryRowsResponse, ResourceCallback, SimpleQueryRowsCallback} from '.';
 import {JobCallback, JobResponse, Table, TableMetadata, TableOptions} from './table';
+import {bigquery} from './types';
```
bcoe: If I'm understanding correctly, we'd already solved this bug in `googleapis`, but were not using the same type definitions in this project?

src/table.ts
```diff
@@ -49,150 +50,65 @@ export interface File {
...
-  maxBadRecords?: number;
-  nullMarker?: string;
-  projectionFields?: string[];
-  quote?: string;
-  schema?: {fields: TableField[]};
-  schemaUpdateOptions?: string[];
-  skipLeadingRows?: number;
-  sourceFormat?: string;
+export type JobLoadMetadata = JobRequest<bigquery.IJobConfigurationLoad>&{
   format?: string;
```
bcoe: I'm relatively new to TypeScript (please excuse any silly questions I have); why do we continue defining `format?`; does this mean that `bigquery.IJobConfigurationLoad` already encompasses all the old types, but we're extending on it to introduce `format`?

src/types.d.ts
```diff
@@ -0,0 +1,2255 @@
...
+     * &quot;CREATE_TABLE&quot;: CREATE [OR REPLACE] TABLE without AS SELECT.
+     * &quot;CREATE_TABLE_AS_SELECT&quot;: CREATE [OR REPLACE] TABLE ... AS
+     * SELECT ... . &quot;DROP_TABLE&quot;: DROP TABLE query.
+     * &quot;CREATE_VIEW&quot;: CREATE [OR REPLACE] VIEW ... AS SELECT ... .
+     * &quot;DROP_VIEW&quot;: DROP VIEW query. &quot;ALTER_TABLE&quot;: ALTER
+     * TABLE query. &quot;ALTER_VIEW&quot;: ALTER VIEW query.
+     */
+    statementType?: string;
+    /**
+     * [Output-only] [Beta] Describes a ti
```
bcoe: I agree with @JustinBeckwith, if this is just a copy of the definition in `googleapis` it would be nice to just use the definitions from the other library -- so that the two libraries don't drift apart.

I'd just create a helper in the `scripts/` folder that copies this file over using `require.resolve` (if I'm understanding the problem correctly).

src/table.ts
callmehiphop: Yep, thats correct. Sometimes we offer convenience options and I believe this is one of them

src/dataset.ts
callmehiphop: Yeah, my best guess was whoever implemented the types initially just looked to see what options we explicitly referenced and as a result they were largely incomplete.

src/types.d.ts
callmehiphop: While I definitely agree if we could just borrow the types, that is the best options. But the apiary types file pulls in dependencies we don't use and contain a number of fields we don't support. 

@JustinBeckwith how are the types generated for apiary? The naming convention makes me think they were generated via the discovery api. Do we see a benefit in writing a cli tool (or something) that generates types based on those responses?

src/dataset.ts
JustinBeckwith: Hey, that was me!  And yea, they ain't great.

src/types.d.ts
JustinBeckwith: They're all generated by the apiary generator:
https://github.com/googleapis/google-api-nodejs-client/blob/master/src/generator/generator.ts

To be honest - I don't really mind if you copy the types out :) Whatever makes sense.

src/dataset.ts
callmehiphop: Commence the public shaming!


Just kidding, I ❤️ you

src/types.d.ts
callmehiphop: Oh nice, I had actually started hacking away at something to create the types. I noticed that the apiary types were missing a few thing

* doesn't look to support `any` types, looks like an empty interface is generated instead (See [JsonValue](https://github.com/googleapis/google-api-nodejs-client/blob/master/src/apis/bigquery/v2.ts#L1631))
* doesn't look to support `additionalProperties` field, also generates an empty interface (See [JsonObject](https://github.com/googleapis/google-api-nodejs-client/blob/master/src/apis/bigquery/v2.ts#L1630)) but should be something like
   ```typescript
    interface JsonObject {
      [key: string]: JsonValue;
    }
   ```

callmehiphop: So just a heads up, I wrote a little CLI tool to generate these via Discovery API. I think there's a couple more things to do before it can be open sourced, but once that happens we can add it as a dev dep and just generate them easily whenever. In the meantime I've pushed another commit that includes the generated types!

callmehiphop:
There were a couple of names that were a little off, since we're going to release this as a major bump I went ahead and changed them.

PR #386, opened 2019-03-17
refactor(samples): fix loadJSONFromGCSTruncate wrong function

steffnay:
- [x] Tests and linter pass


JustinBeckwith:
It looks like we introduced 404's in the docs with the previous samples change.  Would you mind submitting a PR to address those first, so the CI goes green?

PR #387, opened 2019-03-18
fix(job): check for `errorResult` when polling jobs

### LINKED ISSUE #329: Load table from csv file: response error when works fine
#### Environment details

  - OS: Ubuntu 18.04 LTS
  - Node.js version: 8.14.0
  - npm version: 6.4.1
  - `@google-cloud/bigquery` version:@google-cloud/bigquery@1.3.0

#### Steps to reproduce

We start from a CSV that has 2 incorrect lines and we want to load in a bigquery table.

1. When maxBadRecords=0, the job returns an error and the data is not inserted into the table. Expected response. Ok

2. When maxBadRecords=(>=2), the job returns the same error, but the data in the destination is inserted correctly. It is an unexpected response according to the official documentation

[Documentation ](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-csv#bigquery_load_table_gcs_csv-nodejs) says:
maxBadRecords | (Optional) The maximum number of bad records that BigQuery can ignore when running the job. If the number of bad records exceeds this value, an invalid error is returned in the job result. The default value is 0, which requires that all records are valid

Thanks

callmehiphop:
Fixes #329

Would like some confirmation from a BQ team member on this one.

- [x] Tests and linter pass

### DIFF
--- a/test/job.ts
+++ b/test/job.ts
@@ -357,7 +357,8 @@ describe('BigQuery/Job', () => {
       const error = new Error('Error.');
       const apiResponse = {
         status: {
-          errors: error,
+          errorResult: error,
+          errors: [error],
         },
       };
 
@@ -365,7 +366,7 @@ describe('BigQuery/Job', () => {
 
       beforeEach(() => {
         job.getMetadata = (callback: Function) => {
-          callback(null, apiResponse, apiResponse);
+          callback(null, apiResponse);
         };
       });
 
--- a/src/job.ts
+++ b/src/job.ts
@@ -454,28 +454,23 @@ class Job extends Operation {
    * @param {function} callback
    */
   poll_(callback: MetadataCallback): void {
-    this.getMetadata(
-        (err: Error, metadata: Metadata, apiResponse: r.Response) => {
-          // tslint:disable-next-line no-any
-          if (!err && (apiResponse as any).status &&
-              // tslint:disable-next-line no-any
-              (apiResponse as any).status.errors) {
-            // tslint:disable-next-line no-any
-            err = new util.ApiError((apiResponse as any).status);
-          }
+    this.getMetadata((err: Error, metadata: Metadata) => {
+      if (!err && metadata.status && metadata.status.errorResult) {
+        err = new util.ApiError(metadata.status);
+      }
 
-          if (err) {
-            callback(err);
-            return;
-          }
+      if (err) {
+        callback(err);
+        return;
+      }
 
-          if (metadata.status.state !== 'DONE') {
-            callback(null);
-            return;
-          }
+      if (metadata.status.state !== 'DONE') {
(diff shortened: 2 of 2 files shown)


stephenplusplus:
It does seem correct to use `errorResults`, considering these definitions:

### status.errorResults
Final error result of the job. If present, indicates that the job has completed and was unsuccessful. For more information, see troubleshooting errors.

### status.errors[]
The first errors encountered during the running of the job. The final message includes the number of errors that caused the process to stop. Errors here do not necessarily mean that the job has completed or was unsuccessful.

Although, I wonder what type of error could happen in `status.errors[]` that wouldn't be means for the job to be unsuccessful? That seems like it would be more of a warning as opposed to an error.

callmehiphop:
@stephenplusplus so while investigating #329 I learned that you can set a `maxBadRecords` option while uploading a CSV file. Essentially you can have a certain number of erroneous rows and for each error an object will appear in the `errors` array. If you come in under the `maxBadRecords` value then `errorResult` will not be present and the job is considered a success.

test/job.ts
```diff
@@ -357,15 +357,16 @@ describe('BigQuery/Job', () => {
...
+          errors: [error],
         },
       };
 
       const sandbox = sinon.createSandbox();
 
       beforeEach(() => {
         job.getMetadata = (callback: Function) => {
-          callback(null, apiResponse, apiResponse);
+          callback(null, apiResponse);
```
bcoe: nit: it would be cool if we could get a test case around this change in behavior.

callmehiphop: I actually thought this was a behavior change that wasn't pulled over to BigQuery correctly, but now that you said this I think we might have stumbled on to a bug in `nodejs-common`

callmehiphop: Oh, nevermind, my initial guess was right. There was a breaking change in common `0.28.0` and the fix apparently never found its way here.

callmehiphop:
@bcoe somehow I missed your comment, but the response can contain 2 different error fields. `errorResult` and `errors`. As it is today we check for `errors` and assume the job failed. I just learned that `errors` can in some cases serve more as a warning and that `errorResult` is more correct in determining if the job failed. When both `errors` and `errorResult`  are present, the first item in `errors` (array) is actually the same value as `errorResult`, so users will still get the same exact error.

callmehiphop:
@tswast so this error started popping up today

> No matching signature for operator < for argument types: STRUCT<seconds INT64, nanos INT64>, TIMESTAMP. Supported signatures: ANY < ANY

Is there something going on with the backend, or should we preparing a fix for this??

tswast:
This looks like a problem in the back end handling of query parameters, unrelated to this PR. Thanks for the heads up. Investigating.

PR #396, opened 2019-03-21
Release @google-cloud/bigquery v2.1.1

callmehiphop:
This pull request was generated using releasetool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,24 @@
 
 [1]: https://www.npmjs.com/package/@google-cloud/bigquery?activeTab=versions
 
+## v2.1.1
+
+03-21-2019 11:40 PDT
+
+### Bug Fixes
+- fix: correctly encode nested custom date/time parameters ([#393](https://github.com/googleapis/nodejs-bigquery/pull/393))
+- fix(job): check for `errorResult` when polling jobs ([#387](https://github.com/googleapis/nodejs-bigquery/pull/387))
+
+### Documentation
+- docs(samples): adds queryParamsNamed and queryParamsPositional ([#381](https://github.com/googleapis/nodejs-bigquery/pull/381))
+- refactor(samples): fix loadJSONFromGCSTruncate wrong function ([#386](https://github.com/googleapis/nodejs-bigquery/pull/386))
+- refactor(samples): split query and table samples into separate files ([#384](https://github.com/googleapis/nodejs-bigquery/pull/384))
+
+### Internal / Testing Changes
+- fix(tests): update TIMESTAMP param tests ([#394](https://github.com/googleapis/nodejs-bigquery/pull/394))
+- chore: publish to npm using wombat ([#390](https://github.com/googleapis/nodejs-bigquery/pull/390))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,24 @@
 
 [1]: https://www.npmjs.com/package/@google-cloud/bigquery?activeTab=versions
 
+## v2.1.1
+
+03-21-2019 11:40 PDT
+
+### Bug Fixes
+- fix: correctly encode nested custom date/time parameters ([#393](https://github.com/googleapis/nodejs-bigquery/pull/393))
+- fix(job): check for `errorResult` when polling jobs ([#387](https://github.com/googleapis/nodejs-bigquery/pull/387))
```
stephenplusplus: This is a tricky kind of fix, since the interpretation against semver isn't always a clear win for the user. This fix could be breaking change-ish for how the user has handled Jobs in the past.

Before: Job emits 'error'.
After: Job completes successfully.

We don't really know what the user does in each of those scenarios, although it seems relatively safe to fail less for the user.

So I don't know if we should bump the major for this (anyone think we should?), but at least calling it out in the docs might be helpful to someone.

callmehiphop: You make a good point and I believe @bcoe shares a similar opinion. I would be curious to hear what others think as well. While this fix is breaking-ish, it's only because it wasn't implemented correctly to begin with. I think that implies users weren't expecting the code to behave the way it did and that a patch would be appropriate.

bcoe: Personally, I would lean towards calling this a breaking change:

1. if you're making an effort to adhere to semver, it's important to move away from attaching too much meaning to the major digit; changing the major just means you've made a breaking API change, it doesn't need to correlate with an exciting new release of your application.
2. @google-cloud/bigquery has enough users that I'd bet at least one of them was relying on the fact that results would populate even though an error was occurring.

callmehiphop: @bcoe all sounds reasonable to me! I've never had much of a say in our breaking change policies, so I would delegate to @JustinBeckwith on that one.

callmehiphop: FWIW I think this functionality has been broken since [`2.0.4`](https://github.com/googleapis/nodejs-bigquery/releases/tag/v2.0.4) (December) and all body errors have been ignored entirely.

bcoe: Let's compromise and call out the change in the CHANGELOG, as:

```
Note: a bug was fixed that some people may have been relying on, results were still be populated in error
responses when an API returned more than the maximum configured errors. 
```

I think we need to better define our policies around breaking API changes, in relation to `fixes`, and I don't want to block this release.

callmehiphop: This is going to drive some one crazy, but I'm ok trying to make a major release (assuming no one objects) and also trying to squeeze https://github.com/googleapis/nodejs-bigquery/pull/385 in with it.

JustinBeckwith: SGTM.  We closing this out then?

callmehiphop: Yar

PR #397, opened 2019-03-24
feat(geo): add support for geography

steffnay:
Trying to locate which portions of code need to be touched. tests totally failing! 
Just seeking feedback on how this should actually be done.

- [x] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1579,7 +1579,15 @@ paginator.extend(BigQuery, ['getDatasets', 'getJobs']);
  * that a callback is omitted.
  */
 promisifyAll(BigQuery, {
-  exclude: ['dataset', 'date', 'datetime', 'job', 'time', 'timestamp'],
+  exclude: [
+    'dataset',
+    'date',
+    'datetime',
+    'geography',
+    'job',
+    'time',
+    'timestamp',
+  ],
 });
 
 /**
--- a/src/index.ts
+++ b/src/index.ts
@@ -161,7 +161,7 @@ export interface BigQueryDatetimeOptions {
 
 export interface QueryParameter {
   name?: string;
-  parameterType: {type: string;};
+  parameterType: ValueType;
   parameterValue: {arrayValues?: Array<{}>; structValues?: {}; value?: {}};
 }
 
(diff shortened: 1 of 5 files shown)


stephenplusplus:
It looks like the lint task is being a bit faily. You could try running `npm run fix` to sort out the issues and then see if there are any issues left. 🤞

callmehiphop:
We should add some tests to the [SQL Parameter](https://github.com/googleapis/nodejs-bigquery/blob/master/system-test/bigquery.ts#L851) system-tests.

src/index.ts
```diff
@@ -445,6 +445,10 @@ export class BigQuery extends common.Service {
           value = BigQuery.timestamp(new Date(value * 1000));
           break;
         }
+        case 'GEOGRAPHY': {
```
JustinBeckwith: This looks like a feature!  If it is, we should use the PR title `feat(geo): add support for geography` or something like that 😄

system-test/.eslintrc.yml
```diff
@@ -0,0 +1,6 @@
+---
```
JustinBeckwith: Since the sources are all in TypeScript, I don't think we need this file.  I could be wrong though!

test/index.ts
```diff
@@ -572,6 +580,34 @@ describe('BigQuery', () => {
       assert.strictEqual(timestamp.value, EXPECTED_VALUE);
     });
   });
+  // *************
```
JustinBeckwith: What's up with the stars? 🌟

test/index.ts
```diff
@@ -572,6 +580,34 @@ describe('BigQuery', () => {
...
+  // ********
+  // *****
+  // ***
+  // *
+  describe('geography', () => {
+    const INPUT_STRING = 'POINT(1 2)';
+
+    it.skip('should expose static and instance constructors', () => {
```
JustinBeckwith: Why skipped?

steffnay: It's skipped in the similar tests that I used as templates, so I left mine skipped. None of the custom data types' 'should expose static and instance constructors' tests appear to pass when un-skipped. Thoughts?

JustinBeckwith: I am 👎 on having skipped tests anywhere in the code base.  For the ones that were already there ...  I kinda don't know how they got there.   We shouldn't add new skipped ones for sure.  @callmehiphop can you shed some light here?

callmehiphop: @JustinBeckwith I can't, but git blame brought me here https://github.com/googleapis/nodejs-bigquery/pull/227/files#r228039589, so it looks like maybe we meant to deprecate those methods and never did?

JustinBeckwith: We should fire that guy

steffnay: Is there something I should change about the way I implemented geo, or am I able to just remove this test for my PR to be ok before we update the rest?

JustinBeckwith: I think we should just delete the skipped tests.

callmehiphop: I think we should remove that `skip` and see if we can get those tests passing, then open an issue to either remove/fix the other tests in a separate PR

callmehiphop:
@JustinBeckwith @jkwlui can one of you please defeat the cursed cla 🤖 megaboss?

JustinBeckwith:
🎉

PR #407, opened 2019-04-03
fix(types): correct  interface

### LINKED ISSUE #406: Invalid typings of `InsertRowsOptions`
`schema` field in `InsertRowsOptions` is defined as non-optional, whereas it should be required only if `autoCreate` is `true`.

#### Environment details

  - OS: irrelevant
  - Node.js version: irrelevant
  - npm version: irrelevant
  - `@google-cloud/bigquery` version: v3.0.0

#### Steps to reproduce

Try to compile the following using TypeScript compiler:
```ts
import { BigQuery } from '@google-cloud/bigquery';

const bigQuery = new BigQuery();
const table = bigQuery.dataset('dataset').table('table');

table.insert({}, {});
```

merlinnot:
Resolves #406 

- [x] Tests and linter pass

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -56,19 +56,11 @@ export type JobMetadataResponse = [JobMetadata];
 // tslint:disable-next-line no-any
 export type RowMetadata = any;
 
-export type InsertRowsOptionsAutoCreate = {
-  autoCreate?: false;
-  schema?: string | {};
-}|{
-  autoCreate: true;
-  schema: string|{};
-};
-export type InsertRowsOptionsRaw = {
+export type InsertRowsOptions = bigquery.ITableDataInsertAllRequest&{
+  autoCreate?: boolean;
   raw?: boolean;
+  schema?: string|{};
 };
-export type InsertRowsOptions = bigquery.ITableDataInsertAllRequest&
-                                InsertRowsOptionsAutoCreate&
-                                InsertRowsOptionsRaw;
 
 export type InsertRowsResponse =
     [bigquery.ITableDataInsertAllResponse | bigquery.ITable];
--- a/src/table.ts
+++ b/src/table.ts
@@ -56,11 +56,19 @@ export type JobMetadataResponse = [JobMetadata];
 // tslint:disable-next-line no-any
 export type RowMetadata = any;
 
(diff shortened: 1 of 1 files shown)


src/table.ts
```diff
@@ -56,11 +56,19 @@ export type JobMetadataResponse = [JobMetadata];
...
 
-export type InsertRowsOptions = bigquery.ITableDataInsertAllRequest&{
-  autoCreate?: boolean;
-  raw?: boolean;
+export type InsertRowsOptionsAutoCreate = {
+  autoCreate?: false;
+  schema?: string | {};
+}|{
+  autoCreate: true;
   schema: string|{};
```
callmehiphop: Any reason why we don't just make `schema` optional? I like the overloads, but I can't imagine a scenario where `autoCreate` would be set to `false` and `schema` would still be present

merlinnot: It doesn't make sense IMO. I followed the documentation as closely as possible. It does not state that this property cannot be set if `autoCreate` is not `true`.

If there are just two possibilities: `autoCreate: true` + `schema` or none of these, why do we even have two properties? Just for readability? In this case, would you prefer to disallow providing the `schema` if `autoCreate` is not `true`?

callmehiphop: That is a documentation failure on our part then, we completely ignore the `schema` unless `autoCreate: true` is set.

> why do we even have two properties?

That is a good question, `autoCreate` is just a convention used across all the Node.js clients, but as you've pointed out it seems kind of pointless here since `schema` basically implies the same thing. We should open a feature request ticket to merge them into a single option

merlinnot: Ok, I can open an issue. I guess there is a chance to make it a non breaking change if we start ignoring `autoCreate` and start relying on `schema`, or introduce yet another property and keep these two for compatibility, at least for some time. I’d be happy to open an issue and work on a PR once we align on a design.

But for the time being, how do we want to approach fixing current type definitions?

callmehiphop: I'm ok with just marking schema as optional, WDYT?

merlinnot: Done :)

callmehiphop:
@merlinnot thanks for the fix!

PR #414, opened 2019-04-11
refactor(ts): cast arrified rows

callmehiphop:
Looks like the latest `arrify` types are causing an issue for us.

- [x] Tests and linter pass

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -1764,7 +1764,7 @@ class Table extends common.ServiceObject {
         optionsOrCallback :
         cb as InsertRowsCallback;
 
-    rows = arrify(rows);
+    rows = arrify(rows) as RowMetadata[];
 
     if (!rows.length) {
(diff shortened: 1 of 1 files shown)


src/table.ts
```diff
@@ -1775,7 +1775,7 @@ class Table extends common.ServiceObject {
     });
 
     if (!options.raw) {
-      json.rows = arrify(rows).map((row) => {
+      json.rows = rows.map((row: RowMetadata) => {
```
JustinBeckwith: If `rows` is typed, why do you need to be explicit about the type here?

callmehiphop: Interestingly this throws an implicit any, I'm guessing its because `RowMetadata` is actually just an alias to `any`.

callmehiphop:
@bcoe I think this is the result of a patch release that came 3 days ago in `arrify`.

PR #449, opened 2019-05-21
feat(model): dataset model support

callmehiphop:
This adds Dataset Model support.

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -36,7 +36,7 @@ class Model extends common.ServiceObject {
   constructor(dataset: Dataset, id: string) {
     const methods = {
       /**
-       * Delete a model.
+       * Delete the model.
        *
        * @see [Models: delete API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/rest/v2/models/delete}
        *
@@ -55,7 +55,7 @@ class Model extends common.ServiceObject {
        *
        * model.delete((err, apiResponse) => {});
        *
-       * @example <caption>If the callback is omitted we'll return a Promise
+       * @example <caption>If the callback is omitted we'll return a Promise.
        * </caption>
        * const [apiResponse] = await model.delete();
        */
@@ -79,7 +79,7 @@ class Model extends common.ServiceObject {
        *
        * model.exists((err, exists) => {});
        *
-       * @example <caption>If the callback is omitted we'll return a Promise
+       * @example <caption>If the callback is omitted we'll return a Promise.
        * </caption>
        * const [exists] = await model.exists();
        */
@@ -104,18 +104,20 @@ class Model extends common.ServiceObject {
        * const dataset = bigquery.dataset('my-dataset');
        * const model = dataset.model('my-model');
        *
-       * model.get((err, model2, apiResponse) => {
-       *   // `model.metadata` has been populated.
+       * model.get(err => {
+       *   if (!err) {
+       *     // `model.metadata` has been populated.
+       *   }
        * });
        *
-       * @example <caption>If the callback is omitted we'll return a Promise
+       * @example <caption>If the callback is omitted we'll return a Promise.
        * </caption>
-       * const [model2, apiResponse] = await model.get();
+       * await model.get();
        */
       get: true,
 
       /**
-       * Return the metadata associated with the Model.
+       * Return the metadata associated with the model.
        *
        * @see [Models: get API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/rest/v2/models/get}
        *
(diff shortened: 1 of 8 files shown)


callmehiphop:
@tswast @shollyman I think I have the basic implementation of the CRUD methods complete, but I'm not sure how to test it? I tried reviewing https://github.com/googleapis/google-cloud-python/pull/7562 to see how models are created there, but my python isn't so great that I could figure it out 😢. Any hints here would be greatly appreciated!

@sduskis @JustinBeckwith Should I include samples in this PR? Historically some one else has always written them, but it looks like I can pull the region tags from the PRs in other langs.

tswast:
> Should I include samples in this PR? Historically some one else has always written them, but it looks like I can pull the region tags from the PRs in other langs.

Either you or @steffnay should write the samples. I think Steffany wanted to focus on Node.js samples this quarter.

steffnay:
This is great. I can do the samples!

steffnay:
This is great. I can do the samples!

src/model.ts
```diff
@@ -0,0 +1,189 @@
...
+       *
+       * @example
+       * const {BigQuery} = require('@google-cloud/bigquery');
+       * const bigquery = new BigQuery();
+       * const dataset = bigquery.dataset('my-dataset');
+       * const model = dataset.model('my-model');
+       *
+       * model.delete((err, apiResponse) => {});
+       *
+       * @example <caption>If the callback is omitted we'll return a Promise
```
stephenplusplus: Possibly the most nit-picky thing you've heard all year, but some of these sentences end in periods and others don't.

src/model.ts
```diff
@@ -0,0 +1,189 @@
...
+       * @returns {Promise}
+       *
+       * @example
+       * const {BigQuery} = require('@google-cloud/bigquery');
+       * const bigquery = new BigQuery();
+       * const dataset = bigquery.dataset('my-dataset');
+       * const model = dataset.model('my-model');
+       *
+       * const metadata = {
+       *   friendlyName: 'thebestmodelever'
```
stephenplusplus: Can it be "theBestModelEver", or is all lowercase really the deal?

src/model.ts
```diff
@@ -0,0 +1,189 @@
...
+ * const bigquery = new BigQuery();
+ * const dataset = bigquery.dataset('my-dataset');
+ *
+ * const model = dataset.model('my-model');
+ */
+class Model extends common.ServiceObject {
+  constructor(dataset: Dataset, id: string) {
+    const methods = {
+      /**
+       * Delete a model.
```
stephenplusplus: Maybe "Delete this model." since we're in the context of a specific model.

src/model.ts
```diff
@@ -0,0 +1,189 @@
...
+       * @param {object} callback.apiResponse The full API response.
+       * @returns {Promise}
+       *
+       * @example
+       * const {BigQuery} = require('@google-cloud/bigquery');
+       * const bigquery = new BigQuery();
+       * const dataset = bigquery.dataset('my-dataset');
+       * const model = dataset.model('my-model');
+       *
+       * model.get((err, model2, apiResponse) => {
```
stephenplusplus: Can we just show it with the `err` argument, so we don't have to do the `model2` thing? The comment mentions how to get the metadata from the original object, so the natural question people would have is "what is the difference between model and model2? Do I need it?"

src/model.ts
```diff
@@ -0,0 +1,189 @@
...
+       * });
+       *
+       * @example <caption>If the callback is omitted we'll return a Promise
+       * </caption>
+       * const [model2, apiResponse] = await model.get();
+       */
+      get: true,
+
+      /**
+       * Return the metadata associated with the Model.
```
stephenplusplus: Sometimes it says "Model" and others "model." I guess it should always be "model" unless we're talking about a @google-cloud/bigquery "Model" object.

src/model.ts
callmehiphop: Truth be told I just copied a lot of the docs from existing docs, lol.

src/model.ts
callmehiphop: We can do that!

src/model.ts
callmehiphop: Going to follow suit with the other methods and use "Delete the model", implies the same thing I think.

src/model.ts
callmehiphop: Yeah, that example felt kinda weird, I just went that way for completeness.

src/model.ts
stephenplusplus: I figured!

src/model.ts
stephenplusplus: SGTM!

PR #453, opened 2019-05-23
chore: remove unused request types package

JustinBeckwith:


src/table.ts
```diff
@@ -25,7 +25,6 @@ const format = require('string-format-obj');
 import * as fs from 'fs';
 import * as is from 'is';
 import * as path from 'path';
-import * as r from 'request';
```
stephenplusplus: Nothing in the stack was able to catch that we weren't using this?

JustinBeckwith: That is a very good question :) Filed this to track enabling the `no-unused-var` rule in our default tslint config:
https://github.com/google/gts/issues/351

PR #454, opened 2019-05-23
refactor: replaces job#getQueryResults with table#getRows in query method

### LINKED ISSUE #12: bigquery: consider moving from getQueryResults to listTableData
_From @pongad on October 31, 2017 3:35_

The current `query` method uses [`getQueryResults`](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults) RPC to read rows of the results. However, benchmark shows that [`listTableData`](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/list) run significantly faster. Experiments in Go shows `getQueryResults` take 70% longer.

The choice of RPC is an implementation detail and can be changed whenever. However, `listTableData` returns fewer fields. In Java, we decided to tweak the API surface to make it easier to change RPC later. Of course, the change to API needs to happen before GA.

Will Node require any API change?

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2709_

steffnay:
Fixes #12
- [x] Tests and linter pass


src/index.ts
```diff
@@ -1583,8 +1578,16 @@ export class BigQuery extends common.Service {
...
       options = extend({job}, options);
-      job!.getQueryResults(options, callback as QueryRowsCallback);
+      const datasetId = job!.metadata.configuration.query.destinationTable
+        .datasetId;
+      const tableId = job!.metadata.configuration.query.destinationTable
+        .tableId;
+      const dataset = this.dataset(datasetId);
+      const table = dataset.table(tableId);
+
+      table.getRows(options, callback as RowsCallback);
```
callmehiphop: I think this is my mistake, it looks like `getQueryResults` has some polling logic included. I think we just need to do something like

```js
job
  .on('error', err => callback(err))
  .on('complete', () => {
    table.getRows(options, callback);
  });

```

Does that sound right @stephenplusplus?

stephenplusplus: Yes, that sounds right!

steffnay: Thanks! Made this update, still have a failing system-test that I can't locate the fix for.

callmehiphop: Ah, looks like maybe list tabledata won't work for certain queries (like creating models). I'm kind of surprised it returns a table reference at all in this case. We may need to conditionally call `getRows` and `getQueryResults` depending on the type of query.

src/index.ts
```diff
@@ -1583,8 +1582,26 @@ export class BigQuery extends common.Service {
...
+      job!
+        .on('error', err => callback!(err))
+        .on('complete', () => {
+          options = extend({job}, options);
+
+          if (
+            typeof query === 'string' &&
+            (query.includes('CREATE MODEL') || query.includes('REPLACE MODEL'))
+          ) {
+            job!.getQueryResults(options, callback as QueryRowsCallback);
```
callmehiphop: I think this method has polling logic built in, so we can move this check outside of the complete callback and return early.

steffnay: Gotcha

src/index.ts
```diff
@@ -1584,7 +1583,25 @@ export class BigQuery extends common.Service {
       // The Job is important for the `queryAsStream_` method, so a new query
       // isn't created each time results are polled for.
       options = extend({job}, options);
-      job!.getQueryResults(options, callback as QueryRowsCallback);
+      if (
```
stephenplusplus: Could you add a bit of context about why we're doing this extra logic? You could probably just reference the original performance issue & then mention why we need to treat the `[CREATE|REPLACE] MODEL` query types differently.

test/index.ts
```diff
@@ -1782,55 +1773,78 @@ describe('BigQuery', () => {
...
+      };
 
-    it('should assign Job on the options', done => {
-      const fakeJob = {
-        getQueryResults: (options: {}, callback: Function) => {
+      const fakeTable = {
+        getRows(options: {}, cb: Function) {
           assert.deepStrictEqual(options, {job: fakeJob});
-          done();
+          assert.strictEqual(cb, finalCallback);
```
stephenplusplus: We don't really need to simulate executing the callback in this case, since we just want to know that the user's callback was passed to `table.getRows`. In other tests, we just use the `done` function as the simulated user callback, so that it just ends the test:

```js

const fakeTable = {
  getRows(options: {}, cb: Function) {
    assert.deepStrictEqual(options, {job: fakeJob});
    cb(); // done()
  }
};

// ...

bq.query(QUERY_STRING, done);
```

Either way works, if you still want to break out and name the user callback. I would just leave it at the strict comparison of cb === finalCallback in that case, and then just end the test right after.

src/index.ts
steffnay: Done

src/index.ts
```diff
@@ -1590,7 +1589,29 @@ export class BigQuery extends common.Service {
...
       // isn't created each time results are polled for.
       options = extend({job}, options);
-      job!.getQueryResults(options, callback as QueryRowsCallback);
+
+      // table#getRows uses listTableData endpoint, which is a faster method
+      // to read rows of the results. However, it won't work for model queries,
+      // so use the original job#getQueryResults for model queries.
+      if (
+        typeof query === 'string' &&
+        (query.includes('CREATE MODEL') || query.includes('REPLACE MODEL'))
```
callmehiphop: Might want to use a regex here since these could be lowercase. Is there nothing available to us in the response body that would indicate we created a model?

steffnay: Thanks for the catch! Totally meant to do that!

src/index.ts
```diff
@@ -1571,9 +1569,10 @@ export class BigQuery extends common.Service {
     optionsOrCallback?:
       | QueryOptions
       | SimpleQueryRowsCallback
+      | RowsCallback
       | QueryRowsCallback,
-    cb?: SimpleQueryRowsCallback | QueryRowsCallback
-  ): void | Promise<SimpleQueryRowsResponse> | Promise<QueryRowsResponse> {
+    cb?: SimpleQueryRowsCallback | RowsCallback
```
bcoe: It surprised me that `QueryRowsCallback` had to change `RowsCallback`, since in the case of model queries we still return the original response:

```
 job!.getQueryResults(options, callback as QueryRowsCallback);
```

☝️ this is the original slow path right? which we returned as a `QueryRowsCallback`?

Mainly just making sure code-paths return the same shaped object.

test/index.ts
```diff
@@ -1793,7 +1797,39 @@ describe('BigQuery', () => {
       });
     });
 
-    it('should call job#getQueryResults', done => {
+    it('should call table#getRows', done => {
+      const fakeJob = new EventEmitter();
+      // tslint:disable-next-line: no-any
```
bcoe: I recently realized that codecov will aggregate together coverage results from multiple reporters. Along with adding tests (which is awesome) it might be worth changing the `system-test` script in `package.json` to the following:

```
 "system-test": "nyc mocha build/system-test --timeout 600000",
```

Any integration tests that exercise your code path will then also be reflected in codecov 😄

steffnay: Got it. Thanks!

src/index.ts
```diff
@@ -1590,7 +1589,23 @@ export class BigQuery extends common.Service {
       // The Job is important for the `queryAsStream_` method, so a new query
       // isn't created each time results are polled for.
       options = extend({job}, options);
-      job!.getQueryResults(options, callback as QueryRowsCallback);
+
+      // table#getRows uses listTableData endpoint, which is a faster method
+      // to read rows of results.
+
+      job!.getQueryResults(options, (err, rows) => {
```
callmehiphop: So `getQueryResults` is going to get called continuously until the job is done and all the rows have been fetched. That doesn't sound like something we want here, so we'll probably need to set `autoPaginate` to `false`.

src/index.ts
```diff
@@ -1590,7 +1589,23 @@ export class BigQuery extends common.Service {
...
+        if (!err) {
+          if (rows!.length !== 0) {
```
callmehiphop: This throws me off a little bit, does this imply that the best way to go is to poll the `getQueryResults` API until we get a single page of results and only then should we call `listTableData`?

src/index.ts
callmehiphop: I think I pieced this together, I have a feeling we are going to need to manually poll this method until we see the `jobComplete` value set to `true`. If there are 0 rows, we execute the users callback with an empty array and the response object. If a page was returned, then we should call `listTableData`.

@shollyman does that sound right?

src/index.ts
```diff
@@ -1590,7 +1589,23 @@ export class BigQuery extends common.Service {
...
+            const datasetId = job!.metadata.configuration.query.destinationTable
+            .datasetId;
+            const tableId = job!.metadata.configuration.query.destinationTable
+              .tableId;
+            const dataset = this.dataset(datasetId);
+            const table = dataset.table(tableId);
```
callmehiphop: We're probably going to need to cache the `destinationTable` in case users don't want to `autoPaginate` the results

src/index.ts
shollyman: Yeah.  Typical flow for query processing should look something like this in terms of backend interactions:

1. Call bigquery.jobs.insert() to insert the query job into the system, 
   which returns a job resource (and job reference) for future polling
   operations.
   
1. Call bigquery.jobs.getQueryResults to poll (possibly repeatedly) for
   job completion.  Typically you do this with a nonzero timeoutMs param (which
   moves polling to the server side as a hanging get), and zero maxRows
   param (as we're not consuming row data here).

1. When job is complete, you can consume schema from the getQueryResults,
   and then setup the node equivalent of a row iterator if the getQueryResults
   response indicates rows were returned.  Many queries don't return results, 
  such as DML or DDL operations (UPDATE TABLE, ALTER TABLE, etc).
   
1. Row iteration should use bigquery.tabledata.list to fetch pages of row data. 
  Each page optionally returns a next page token, which is passed to the next 
  invocation to advance the cursor (don't use explicit row offsets) until no further
  page token is provided.

src/index.ts
steffnay: Yeah, that's what Seth told me they do in the other languages and how I should handle it to be consistent.

PR #460, opened 2019-05-28
docs(samples): adds get, delete, list, update model samples

steffnay:
- [x] Tests and linter pass

### DIFF
--- a/samples/deleteModel.js
+++ b/samples/deleteModel.js
@@ -16,6 +16,11 @@
 
 'use strict';
 
+// sample-metadata:
+//   title: BigQuery Delete Model
+//   description: Deletes a model.
+//   usage: node deleteModel.js <DATASET_ID> <MODEL_ID>
+
 function main(datasetId, modelId) {
   // [START bigquery_delete_model]
   // Import the Google Cloud client library
--- a/samples/deleteModel.js
+++ b/samples/deleteModel.js
@@ -1,0 +1,45 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+function main(datasetId, modelId) {
+  // [START bigquery_delete_model]
+  // Import the Google Cloud client library
+  const {BigQuery} = require('@google-cloud/bigquery');
+
+  async function deleteModel() {
+    // Deletes a model named "my_existing_model" in "my_dataset".
+
+    /**
+     * TODO(developer): Uncomment the following lines before running the sample
+     */
+    // const datasetId = "my_dataset";
+    // const modelId = "my_model";
+
+    // Create a client
+    const bigqueryClient = new BigQuery();
+
+    const dataset = bigqueryClient.dataset(datasetId);
+    const model = dataset.model(modelId)
+    const [apiResponse] = await model.delete();
+
+    console.log(`Model ${modelId} deleted.`);
+  }
+  deleteModel();
+  // [END bigquery_delete_model]
+}
+main(...process.argv.slice(2));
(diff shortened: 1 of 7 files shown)


samples/package.json
```diff
@@ -12,7 +12,7 @@
     "node": ">=8"
   },
   "scripts": {
-    "test": "mocha --timeout 60000"
+    "test": "mocha --timeout 600000"
```
leahecole: Does this need to be this high?

samples/test/models.test.js
```diff
@@ -0,0 +1,150 @@
...
+const cp = require('child_process');
+const uuid = require('uuid');
+
+const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
+
+const bigquery = new BigQuery();
+
+describe(`Models`, () => {
+  const datasetId = `gcloud_tests_${uuid.v4()}`.replace(/-/gi, '_');
+  const modelId = `gcloud_tests_${uuid.v4()}`.replace(/-/gi, '_');
```
leahecole: Are these two variables exactly the same? If so, should there be something distinguishing which is the dataset and which is the model? If that distinguishing factor happens later, consider reducing this to a single var (unless I'm missing something and they actually are different)

samples/test/models.test.js
```diff
@@ -0,0 +1,150 @@
...
+  after(async () => {
+    await bigquery
+      .dataset(datasetId)
+      .delete({force: true})
+      .catch(console.warn);
+  });
+
+  it(`should retrieve a model if it exists`, async () => {
+    const output = execSync(`node getModel.js ${datasetId} ${modelId}`);
+    assert.match(output, /Model:/);
```
leahecole: Can/should these assertions be combined into a single one?

samples/test/models.test.js
```diff
@@ -0,0 +1,150 @@
...
+    assert.match(output, new RegExp(datasetId));
+    assert.match(output, new RegExp(modelId));
+  });
+
+  it(`should list models`, async () => {
+    const output = execSync(`node listModels.js ${datasetId}`);
+    assert.match(output, /Models:/);
```
leahecole: Can/should these assertions be combined into a single one?

samples/test/models.test.js
```diff
@@ -0,0 +1,150 @@
...
+    assert.match(
+      output,
+      new RegExp(`${modelId} description: A really great model.`)
+    );
+  });
+});
+
+describe(`Delete Model`, () => {
+  const datasetId = `gcloud_tests_${uuid.v4()}`.replace(/-/gi, '_');
+  const modelId = `gcloud_tests_${uuid.v4()}`.replace(/-/gi, '_');
```
leahecole: Are these two variables exactly the same? If so, should there be something distinguishing which is the dataset and which is the model? If that distinguishing factor happens later, consider reducing this to a single var (unless I'm missing something and they actually are different)

samples/test/models.test.js
```diff
@@ -0,0 +1,150 @@
...
+    await bigquery.createDataset(datasetId, datasetOptions);
+
+    // Run query to create a model
+    const [job] = await bigquery.createQueryJob(queryOptions);
+
+    // Wait for the query to finish
+    await job.getQueryResults();
+  });
+
+  after(async () => {
```
leahecole: If the delete model test fails, will the model be cleaned up? Does that matter?

samples/test/models.test.js
callmehiphop: `uuid.v4()` creates a unique ID, so they should be different.

samples/test/models.test.js
callmehiphop: Hooks run regardless of whether or not a test passes or fails. Since models are tied to datasets, this hooks should clean up any models created in this spec.

samples/deleteModel.js
```diff
@@ -0,0 +1,45 @@
...
+    const dataset = bigqueryClient.dataset(datasetId);
+    const model = dataset.model(modelId);
+    await model.delete();
+
+    console.log(`Model ${modelId} deleted.`);
+  }
+  deleteModel();
+  // [END bigquery_delete_model]
+}
+main(...process.argv.slice(2));
```
callmehiphop: I'm kind of surprised to see we don't catch and print errors here, I think they'd result in a unhandled rejection otherwise? Looks like this is true to all the BQ samples. @JustinBeckwith should we open an issue for this?

samples/test/models.test.js
leahecole: LOL yep I see it now. I'm gonna go ahead and blame this one on jetlag to save face 🙈

samples/deleteModel.js
JustinBeckwith: I think we have a little confusion around the state of the art here.  Mind opening a discussion issue in googleapis/google-cloud-node?

PR #464, opened 2019-05-30
fix!: store job id instead of job instance

### LINKED ISSUE #463: Queries from the Node.js BigQuery client v2.0.0 and higher are not returning results
Running a query in a Node.js 8 environment using the BigQuery Client Library, v3.0.0 results in an empty response returned to the caller.

Reproduction steps and observed behavior: 

Sample query: "with foo as (select 'Travis' as name) select count(*) as total from foo"

Expected behavior:  The query() method should return COUNT = 1 after running the aggregate query from above.  

Actual behavior: It's currently returning empty for bq client 2.0.0 and above.

callmehiphop:
Fixes #463

- [x] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -94,7 +94,7 @@ export type Query = JobRequest<bigquery.IJobConfigurationQuery> & {
   params?: any[] | {[param: string]: any};
   dryRun?: boolean;
   defaultDataset?: Dataset;
-  job?: Job;
+  job?: string;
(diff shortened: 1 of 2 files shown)


callmehiphop:
@cmath if you want to give this a test drive to see if it resolves your issue, that would be awesome!

callmehiphop:
Closing in favor of #465 which is not a breaking change.

PR #484, opened 2019-06-17
refactor(query)!: use listTableData instead of getQueryResults

### LINKED ISSUE #12: bigquery: consider moving from getQueryResults to listTableData

callmehiphop:
Fixes #12

- [x] Tests and linter pass


callmehiphop:
@stephenplusplus @shollyman looking for some early(ish) feedback on the implementation before I get too far into tests.

src/index.ts
```diff
@@ -1568,44 +1565,61 @@ export class BigQuery extends common.Service {
...
-   * This method will be called by `createQueryStream()`. It is required to
-   * properly set the `autoPaginate` option value.
-   *
    * @private
    */
-  queryAsStream_(query: Query, callback?: SimpleQueryRowsCallback) {
-    if (query.job) {
-      query.job.getQueryResults(query, callback as QueryRowsCallback);
-      return;
+  async runQuery_(query: Query, options: QueryOptions): Promise<RowsResponse> {
```
callmehiphop: @shollyman a second pair of eyes on this method would be great, basically it does the following

1. creates a query job
2. polls `getQueryResults` until `jobComplete` is `true` (should we set `maxResults` to something like 0 or 1?)
3. upon completion, if rows are found then we call `listTableData` (aka `table.getRows()`) via returned `destinationTable`

src/index.ts
```diff
@@ -1568,44 +1565,61 @@ export class BigQuery extends common.Service {
...
+      return [rows, null, results!];
     }
-    this.query(query, {autoPaginate: false}, callback);
+
+    const {datasetId, tableId} = resp.configuration!.query!.destinationTable!;
+    const table = this.dataset(datasetId!).table(tableId!);
+
+    table.metadata = {schema: results!.schema};
+
+    const tableData = await table.getRows(options);
```
callmehiphop: All our promises return tuples, so `tableData` is essentially `[rows, nextPageToken, fullResponseBody]`

src/index.ts
tswast: > should we set maxResults to something like 0 or 1?

We set it to 0 in Python.

https://github.com/googleapis/google-cloud-python/blob/387b8e5de88d6af58b45e513426788c36a733fab/bigquery/google/cloud/bigquery/client.py#L857

src/index.ts
shollyman: It looks like for query dryrun this returns the job insert response as the third element, but for real queries the third element is the getQueryResults response?  This seems inconsistent, but I may be misunderstanding.

callmehiphop: No, that is exactly what you're seeing. This is a pattern you'll see in pretty much every Node client library, we always return the full response body as the last element. I agree that it can be a bit odd in methods that make several API requests.

@bcoe any feelings here?

src/index.ts
callmehiphop: Excellent, thanks @tswast!

src/index.ts
steffnay: @callmehiphop @shollyman @bcoe  Any progress on this? It's an issue that I'm getting pinged on regarding the status.

bcoe: no super strong opinion, as long as we're consistent; If we had a do over, I'd probably return an object with meaningful keys names, rather than a tuple, which one could then destructure -- but I don't want to break our design pattern.

src/index.ts
```diff
@@ -1568,44 +1565,81 @@ export class BigQuery extends common.Service {
...
+
+    const queryOptions = Object.assign({}, options, {
+      autoPaginate: false,
+      maxResults: 0,
+    });
+
+    let rows, nextQuery, results;
+
+    do {
+      [rows, nextQuery, results] = await job.getQueryResults(queryOptions);
```
callmehiphop: @tswast does the Python client use any kind of backoff in between `getQueryResults` calls?

tswast: Yes, in two ways.

1. The default retry exponential backoff for each individual call:

   https://github.com/googleapis/google-cloud-python/blob/master/bigquery/google/cloud/bigquery/job.py#L2793-L2829
  
   Note: we set a timeout of 10 seconds per call. The call hangs until timeout or the job completes.
2. `result()` checks the value of `done()` until it's true or the call times out.

   https://github.com/googleapis/google-cloud-python/blob/83f0262fc52bcf63b57e3902d83656c70a7cbcc2///api_core/google/api_core/future/polling.py#L87-L105

callmehiphop:
I recently got to a point where I thought the code was ready, but found that the performance actually took a hit. This threw me off a bit but after debugging I've found a perf related issue in one of our dependencies. I've opened an issue (https://github.com/googleapis/nodejs-paginator/issues/134) and I don't think we should move forward with this PR until it is resolved.

callmehiphop:
I think we should be good here if anyone wants to take a final look through before merging 😄

src/index.ts
stephenplusplus: @callmehiphop do we currently, or need to, do any of that here?

src/index.ts
```diff
@@ -1568,44 +1565,81 @@ export class BigQuery extends common.Service {
...
-      }
-      // The Job is important for the `queryAsStream_` method, so a new query
-      // isn't created each time results are polled for.
-      options = extend({job}, options);
-      job!.getQueryResults(options, callback as QueryRowsCallback);
-    });
+
+    query = typeof query === 'string' ? {query} : query;
+
+    if (query.table) {
```
stephenplusplus: Is `table` always a property we set, or is it possible the user can provide the string name of the table they want to query from? (I had a hard time following the TypeScript definition trail to determine what options `query` has)

src/index.ts
```diff
@@ -1568,44 +1565,81 @@ export class BigQuery extends common.Service {
...
+      maxResults: 0,
+    });
+
+    let rows, nextQuery, results;
+
+    do {
+      [rows, nextQuery, results] = await job.getQueryResults(queryOptions);
+    } while (!results!.jobComplete);
+
+    if (!Number(results!.totalRows)) {
```
stephenplusplus: What does this check? Maybe there's a less exciting! way to write it?

callmehiphop: Gives back the number of rows inserted into the destination table. We need this for working with the Model API since models are created via SQL queries and will never return rows. They still create a destination table and querying against it results in an error.

src/index.ts
callmehiphop: It is something we set, it actually is a `Table` reference (not just the name) that we use to make additional queries against.

src/index.ts
callmehiphop: The default retry is baked in from the common libraries, but we could probably do a better job of implementing step 2.

callmehiphop:
Technically this will be a TypeScript breaking change, so I'm going to hold off on merging until a major bump presents itself.

PR #488, opened 2019-06-26
docs(samples): Adds samples

steffnay:
- [x] Tests and linter pass


samples/cancelJob.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+'use strict';
+
+function main(jobId) {
+  // [START bigquery_cancel_job]
+  // Import the Google Cloud client library
+  const {BigQuery} = require('@google-cloud/bigquery');
+
+  async function cancelJob(jobId = 'existing-job-id') {
```
JustinBeckwith: Instead of accepting the `jobId` as a parameter here, we should rely on the parameter getting passed into the `main` method above.  Then we should have one of those:

```js
// TODO(developer): uncomment the following lines
// const jobId = 'existing-job-id';
```

... type of thingies.

samples/cancelJob.js
```diff
@@ -0,0 +1,41 @@
...
+    // Attempt to cancel a job.
+
+    // Create a client
+    const bigqueryClient = new BigQuery();
```
JustinBeckwith: Let's just call this `client`.  Also - it's super important to instantiate this *outside* of the function, and right after the `require` statement.  It's important because we want users to create this object *once*, and re-use the client.  There's all kind of caching that happens inside of that object, so if they create these in a loop, or ever time a function is called, it's hell on performance.

samples/cancelJob.js
```diff
@@ -0,0 +1,41 @@
...
+    // Create a job reference
+    const job = bigqueryClient.job(jobId);
+
+    // Attempt to cancel job
+    const [apiResult] = await job.cancel();
+
+    console.log(apiResult.job.status);
+  }
+  // [END bigquery_cancel_job]
+  cancelJob(jobId);
```
JustinBeckwith: as mentioned above, let's drop passing the `jobId` parameter here

samples/copyTableMultipleSource.js
```diff
@@ -0,0 +1,52 @@
...
+
+'use strict';
+
+function main(datasetId, sourceTable, destinationTable) {
+  // [START bigquery_copy_table_multiple_source]
+  // Import the Google Cloud client library
+  const {BigQuery} = require('@google-cloud/bigquery');
+
+  async function copyTableMultipleSource(
+    datasetId = 'my_dataset', // Existing dataset
```
JustinBeckwith: All the same feedback as above just here 🙃

samples/test/tables.test.js
```diff
@@ -272,6 +319,20 @@ describe('Tables', () => {
     assert.match(output, /Inserted 2 rows/);
   });
 
+  it(`copy multiple source tables to a given destination`, async () => {
+    execSync(`node createTable.js ${datasetId} destinationTable "${schema}"`);
+    const output = execSync(
+      `node copyTableMultipleSource.js ${datasetId} ${tableId} destinationTable`
+    );
+    assert.include(
```
JustinBeckwith: I don't think this does what you want it to 🙃  This statement:

```js
'sourceTable' && 'destinationTable' && 'createDisposition' && 'writeDisposition'
```

Will always evaluate to `writeDisposition`.  JavaScript does weird truthy checks with `&&` that leads to this - try it in the repl!

You probably want to write 4 separate `assert.include` statements

samples/cancelJob.js
steffnay: @alixhami had a point that people miss the commented out variables so it's easier to see what to change with the input. I really have no opinion either way, just want to know if there is some standard so I can be consistent.

JustinBeckwith: Ah, understood.  The usage of that pattern is codified here:
https://g3doc.corp.google.com/company/teams/cloud-devrel/dpe/samples/nodejs.md?cl=head

If we want to change it - it's something we'd have to all talk about, agree on, plan a large scale change, etc.  I want to be really careful about following the stuff in the rubric.

samples/cancelJob.js
steffnay: many of the older samples call it bigquery rather than client or bigqueryClient. Does bigquery work or should I transition to client?

samples/cancelJob.js
alixhami: It's a bit confusing between languages, because the rubric committee's recommendation was default parameters, not commented out variables. Then, some languages deviated from that once the per-language sample standard pages were created.

JustinBeckwith: Ah, so the good news here is that we do both!  Please do keep the default parameters, just move em up to the `main` function.  

The default parameters are for folks who clone the repo and run the samples locally.  The commented out declarations are for folks browsing cloud.google.com.  Those folks won't see the `main` method in the snippet ... so commented out we go :)

JustinBeckwith: (side note - would be a great topic to start an internal thread on so we can all get on the same page)

samples/cancelJob.js
JustinBeckwith: Honestly either/or is fine by me

samples/cancelJob.js
steffnay: Thanks both!!

samples/test/tables.test.js
steffnay: Gotcha.

PR #490, opened 2019-07-04
feat: adds method to list projects

### LINKED ISSUE #84: Support List Projects
Hi,

Would be nice if the library could have a method to list the projects. 

- Cf. https://cloud.google.com/bigquery/docs/reference/rest/v2/projects/list

Regards

steffnay:
Fixes #84 

- [x] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -120,6 +120,8 @@ export type DatasetsCallback = PagedCallback<
 export type DatasetResponse = [Dataset, bigquery.IDataset];
 export type DatasetCallback = ResourceCallback<Dataset, bigquery.IDataset>;
 
+export type GetProjectsOptions = PagedRequest<bigquery.projects.IListParams>;
+
 export type GetJobsOptions = PagedRequest<bigquery.jobs.IListParams>;
 export type GetJobsResponse = PagedResponse<
   Job,
@@ -128,6 +130,12 @@ export type GetJobsResponse = PagedResponse<
 >;
 export type GetJobsCallback = PagedCallback<
   Job,
+  GetJobsOptions,
+  bigquery.IJobList
+>;
+ 
+export type GetProjectsCallback = PagedCallback<
+  Job, // PROJECT???
   GetJobsOptions,
   bigquery.IJobList
 >;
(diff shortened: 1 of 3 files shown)


test/index.ts
```diff
@@ -1856,4 +1871,126 @@ describe('BigQuery', () => {
       bq.queryAsStream_(query, done);
     });
   });
+
+  describe('listProjects', () => {
```
bcoe: one question, this seems really close to the structure of `getJobs` and `getDatasets`, but in this case we're using the prefix `list` rather than `get`; wondering if there's a precedent for this naming elsewhere?

tswast: `list` is the one platform verb name. Makes sense to stay consistent with the other resources, though.

jkwlui: We've been using `getXs` in nodejs-storage as well - even though the API is actually `list`.

stephenplusplus: Just providing some additional context that was shared in the Hangout:

We differentiated our `getX` methods from the upstream `list` operations, as ours return custom, functional objects,  e.g. `storage.getBuckets()` returns `Bucket`s.

In our opinion at the time, `storage.listBuckets()` wouldn't communicate you'd be getting such objects, but instead only a "list" of strings or barebones descriptors. They are two different behaviors, and the one that returns functional objects is more useful, as it does everything `list` does and more.

In the case of this new method, however, `getProjects()` doesn't return a `Project` object. It actually is a list of barebones descriptors. I don't want to overcomplicate something that doesn't need it, but I thought of a few ways we could handle this:

1. Move this behavior to the common or auth library. As a comparison, the auth library exposes a `getProjectId()` method, and we re-expose it through common. This couples nicely, because a project ID can come from a key file, and is usually required to figure out auth permissions. Similarly, the projects you are allowed to oversee is an auth matter, as opposed to something that is unique to BigQuery or any single library.

2. Keep this method local, but rename it e.g. `getProjectIds()`. This follows the conventional `getX` terminology, but also doesn't imply we will be returning an object.

If 1. sounds interesting, it could be worth considering depending on the Resource Manager API, `@google-cloud/resource`, as it is quite a thin wrapper around Project behavior, including `getProjects()`, where it actually does return a `Project` object that has its own behavior.

steffnay: This idea makes sense to me. @callmehiphop & @shollyman, do either of you see a problem with this?

callmehiphop: I think it's fine to call it `getProjects` personally.

@stephenplusplus there's a lot more in the [response](https://cloud.google.com/bigquery/docs/reference/rest/v2/projects/list#response-body) than just the project ID, so `getProjectIds` doesn't really make sense to me unless we plan on mapping the response to exclude the other fields, which honestly I don't think we should do. On the flip side the only method available to us is list, so it doesn't make sense to create an actual `Project` class either.

stephenplusplus: I didn't even see that it was actually a BigQuery API endpoint. Does anyone know if it is just a map to https://cloud.google.com/resource-manager/reference/rest/v1/projects/list?

shollyman: So, a bit of a dissenting view here.  I'm generally in the "don't implement this at all" camp for this particular BQ method.

The `bigquery.projects.list` API is a utility method from the eons before the project service was available externally, and existed to populate the dropdown in the bigquery classic UI.  What it does is use resource manager to enumerate projects where the user has some level of project access, and uses service usage to filter that list down to projects where the BigQuery service is enabled.  Now that those services are externalized we should likely direct users to leverage those mechanisms, as they're much more functionally useful.  Poking around I see https://github.com/googleapis/nodejs-resource/tree/master/samples#list-projects is already a sample for resource manager enumeration, but I don't see an example of talking to service usage.

steffnay: Okay, so should I just close this?

shollyman: I'm now pushing through changes to the generated API docs to make it clearer that resource manager is the better choice for this data.  Unless you've got a particular user demanding this, I'd go ahead and drop it.

shollyman: https://cloud.google.com/bigquery/docs/reference/rest/v2/projects/list now indicates resource manager as the alternative.  It'll make it's way into discovery and generated artifacts over time as releases progress.

src/index.ts
```diff
@@ -1441,6 +1456,75 @@ export class BigQuery extends common.Service {
...
+   * //-
+   * bigquery.getProjects().then(function(data) {
+   *   const projects = data[0];
+   * });
+   */
+  getProjects(
+    optionsOrCallback?: GetProjectsOptions | GetProjectsCallback,
+    cb?: GetProjectsCallback
+  ): void | Promise<GetProjectsResponse> {
+    const that = this;
```
AVaksman: Nit -> seems `that` is not used in this method.

PR #501, opened 2019-07-17
test(samples): deflake table sample tests

### LINKED ISSUE #500: deflake sample tests
Not sure what happened, but I've been seeing a lot of sample failures the last few days.

callmehiphop:
Fixes #500

Looks like we are getting rate limited on a test that uploads a CSV file, which in turn causes a series of tests that depend on said file to fail.

### DIFF
--- a/samples/test/tables.test.js
+++ b/samples/test/tables.test.js
@@ -57,6 +57,9 @@ describe('Tables', () => {
     ]);
   });
 
+  // to avoid getting rate limited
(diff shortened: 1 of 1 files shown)


callmehiphop:
@bcoe @steffnay I'm not a huge fan of adding a delay in between tests, but I think its unavoidable here. Any thoughts?

JustinBeckwith (approved):

PR #518, opened 2019-08-08
fix(types): update to the latest discovery types

callmehiphop:
Some of the Model response objects have changed upstream, this is just us generating updated types for the request/response types.

- [ ] Tests and linter pass

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -3,488 +3,13 @@
  */
 declare namespace bigquery {
   /**
-   * Information about a single training query run for the model.
-   */
-  type ITrainingRun = {
-    /**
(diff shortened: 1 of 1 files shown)


callmehiphop:
@bcoe @JustinBeckwith is there anyway we could leverage Kokoro to generate the types for us? 

/cc @shollyman

shollyman:
Is types.d.ts fully autogenerated, or partially?  It's not clear from this PR how that happens.

callmehiphop:
@shollyman we have an [npm script](https://github.com/googleapis/nodejs-bigquery/blob/master/package.json#L48) that generates that file when ran.

PR #528, opened 2019-09-03
feat(typescript): generate latest request/response types

### LINKED ISSUE #525: bigquery:  implement support use_avro_logical_types for export jobs
With the next discovery release, BigQuery will support control of logical type annotations when specifying export jobs that use the Apache Avro file format.  This will be controlled via the use_avro_logical_types.

Description from the API layer: 

[Optional] If destinationFormat is set to "AVRO", this flag indicates
whether to enable extracting applicable column types (such as TIMESTAMP) to
their corresponding AVRO logical types (timestamp-micros), instead of only
using their raw types (avro-long).

This should mirror reciprocal support in load jobs for specifying whether logical type annotations should be used when building the schema of a table.

callmehiphop:
Fixes #525

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -33,17 +33,10 @@ declare namespace bigquery {
 
   type IBqmlTrainingRun = {
     /**
-     * [Output-only, Beta] List of each iteration results.
-     */
-    iterationResults?: Array<IBqmlIterationResult>;
-    /**
-     * [Output-only, Beta] Training run start time in milliseconds since the epoch.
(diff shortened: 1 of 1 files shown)


callmehiphop:
@bcoe so about the license check failing, I imagine this is also an issue with any grpc/typescript libraries where we generate types from the proto files. Doesn't look like there's a solution in place for it yet though, any thoughts?

JustinBeckwith (approved):

PR #545, opened 2019-09-18
feat: implemented top level method for encode

### LINKED ISSUE #19: FR: Let clients use encodeValue function
I'd really like to do this:

```ts
const bq = BigQuery({ projectId: 'prjctd' })
const table = bq.dataset('dtst').table('tbl')
const valueToInsert = {
  insertId: 'myInsertId',
  json: table.encodeValue_(myItem)
}
```

but currently the function is not visible. I can copy it to my project for sure, but it would be nicer to be able to use it.

praveenqlogic01:
fixes #19 
impletmented top level method for encode function

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -363,6 +363,17 @@ export class BigQuery extends common.Service {
   }
 
   /**
+   * Convert a row entry from native types to their encoded types that the API
+   * expects.
+   *
+   * @param {*} value The value to be converted.
+   * @returns {*} The converted value.
+   */
+  static encode(value?: {} | null): {} | null {
+    return Table.encodeValue_(value);
+  }
+
+  /**
    * Merge a rowset returned from the API with a table schema.
    *
    * @private
--- a/test/index.ts
+++ b/test/index.ts
@@ -1137,6 +1137,101 @@ describe('BigQuery', () => {
     });
   });
 
+  describe('encode', () => {
+    it('should return null if null, undefined or string', () => {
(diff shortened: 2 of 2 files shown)


callmehiphop:
This looks like a good start to me!

Part of the user request was to be able to pass in already encoded data. While this change does allow the user to manually encode their data, we don't actually provide a way for them to pass it to the client. I think in addition to this we'll probably want to accept a new option anywhere we call `Table.encodeValue_` as to prevent encoding already encoded data.

I'm just spitballing, but I think it could look something like this

```js
const {BigQuery} = require('@google-cloud/bigquery');

const table = bigquery.dataset('my-dataset').table('my-table');

const data = BigQuery.encode({ ... });

table.insert(data, {encode: false}, (err, resp) => {
  // ...
});
```

@shollyman @tswast @stephenplusplus any thoughts or feelings on this?

bcoe:
@stephenplusplus any thoughts about this PR? would be nice to unblock it.

tswast:
I wrote some thoughts in the issue: https://github.com/googleapis/nodejs-bigquery/issues/19#issuecomment-537117818

The more I work on encoding logic in the Python client, the more I realize we need the table schema to disambiguate types in many cases.

bcoe:
@praveenqlogic, sounds like you should sync up with @tswast on this specific issue and we can try to come up with a holistic approach to encoding.

stephenplusplus:
I went back to #19 to investigate the use cases for exposing a top level encode(), and we discovered we didn't really have one 🤷‍♂.

PR #548, opened 2019-09-21
refactor(types): refactors index to export all types

### LINKED ISSUE #456: Can't Import Useful Types From Main Declaration File
I am a little stumped about how I should be importing types that aren't exported directly from `index.d.ts`. 

For example, updating to `@google-cloud/bigquery 4.0` broke this code in my project:

custom type file
```typescript
import { Job } from '@google-cloud/bigquery';

interface IBigQueryJobs {
  [tableName: string]: Job;
}

export interface ITablesToBigQueryJobState {
  bigQueryJobs: IBigQueryJobs;
}
```
Script file
```typescript
const jobState: ITablesToBigQueryJobState = {
  bigQueryJobs: {},
};
// ... 
const [result] = await bqTable.load(filePath, loadJobMetaData);
// ... 
jobState.bigQueryJobs[tableName] = result;
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
type error:
```
Type 'IJob' is missing the following properties from type 'Job': bigQuery, getQueryResultsStream, cancel, getQueryResults, and 33 more.
```
I looked for the `IJob` type, or an equivalent, from the main declaration file, and could not find one. So finally I just imported directly from `'@google-cloud/bigquery/build/src/types'` which seemed wrong, though it did fix the type error:
custom type file
```typescript
import bigquery from '@google-cloud/bigquery/build/src/types';

interface IBigQueryJobs {
  [tableName: string]: bigquery.IJob;
}

export interface ITablesToBigQueryJobState {
  // ...
  bigQueryJobs: IBigQueryJobs;
}
```

Is there another way I should be thinking about this? Thank you! 

#### Environment details

  - OS: `10.14.3`
  - Node.js version: `v8.12.0`
  - npm version: `6.5.0`
  - `@google-cloud/bigquery` version: `"4.0.0"`

steffnay:
Fixes #456 

- [x] Tests and linter pass


bcoe:
@stephenplusplus I believe @callmehiphop might be away for a little bit? mind providing some review.

src/job.ts
```diff
@@ -35,7 +35,7 @@ import {
   QueryRowsCallback,
   QueryRowsResponse,
   RequestCallback,
-} from '../src';
+} from '../src/bigquery';
```
callmehiphop: Can this just be `./bigquery`?

src/table.ts
```diff
@@ -40,7 +40,7 @@ import {
   PagedCallback,
   JobRequest,
   PagedRequest,
-} from '../src';
+} from '../src/bigquery';
```
callmehiphop: Same question as previous comment

src/bigquery.ts
```diff
@@ -0,0 +1,1776 @@
...
+}
+
+/**
+ * {@link Dataset} class.
+ *
+ * @name BigQuery.Dataset
+ * @see Dataset
+ * @type {constructor}
+ */
+export {Dataset};
```
callmehiphop: We could probably get rid of all these exports since the index file should be importing/exporting from their actual source files now.

test/bigquery.ts
```diff
@@ -30,9 +30,17 @@ import * as proxyquire from 'proxyquire';
...
+  BigQueryDate,
+  Dataset,
+  Job,
+  Table,
+  JobOptions,
+  TableField,
+} from '../src';
+// import {Dataset} from '../src/dataset';
+// import {JobOptions, Job} from '../src/job';
+// import {TableField, Table} from '../src/table';
```
callmehiphop: Lets delete these comments

steffnay: thanks for spotting that!

PR #566, opened 2019-11-03
docs(samples): adds user authentication samples

steffnay:


samples/authUserFlow.js
```diff
@@ -0,0 +1,78 @@
...
+  // [END bigquery_auth_user_flow]
+  // [START bigquery_auth_user_query]
+  async function query() {
+    const {BigQuery} = require('@google-cloud/bigquery');
+
+    const bigquery = new BigQuery({
+      projectId: projectId,
+    });
+
+    bigquery.authClient.cachedCredential = refreshClient;
```
stephenplusplus: I believe we should support passing an auth client instance right to the BigQuery constructor:

```js
const bigquery = new BigQuery({
  authClient: refreshClient,
  projectId: projectId,
});
```

The common library is ready for this: https://github.com/googleapis/nodejs-common/blob/20f614a637a9573abe59b9fe6ec45e45cb59d03e/src/service.ts#L61, although it doesn't seem like the BigQuery constructor is allowing it to be done: https://github.com/googleapis/nodejs-bigquery/blob/6729fe8f0a13e660a289c641aae3b21411ed974e/src/bigquery.ts#L259 (it should be in `config`).

samples/authUserFlow.js
```diff
@@ -0,0 +1,78 @@
...
+  // const refreshToken = 'REFRESH TOKEN'
+  // const projectId = 'PROJECT ID'
+  const credentials = {
+    type: 'authorized_user',
+    client_id: clientId,
+    client_secret: clientSecret,
+    refresh_token: refreshToken,
+  };
+
+  const refreshClient = new UserRefreshClient();
```
stephenplusplus: Is this not automatically done if you pass in the `credentials` object above as:

```js
new BigQuery({credentials});
```

?

steffnay: @stephenplusplus yep, this totally works.

samples/authUserFlow.js
steffnay: I updated it to just passing the credentials to the constructor, per your later comment

stephenplusplus:
LGTM, although I'm not sure the testing philosophy here. (cc @callmehiphop)

Do we need a new test for this? I know the query used in this sample is repeated throughout various tests. I'm not sure if the tests are repeated for each instance as well.

I assume the things we would want to test about this are:

  1. Is this type of authentication valid
  1. Does the code that runs the query and returns the results work as expected

We can copy and paste the test for #&shy;2, but #&shy;1 is tricky. Would we have to intercept the auth library to make sure the `credentials` object was received, but return a normal auth instance that the other tests usually run with?

samples/package.json
```diff
@@ -17,6 +17,7 @@
   "dependencies": {
     "@google-cloud/bigquery": "^4.2.1",
     "@google-cloud/storage": "^4.0.0",
+    "google-auth-library": "^5.5.1",
```
callmehiphop: Why do we need this as a dependency?

samples/authUserFlow.js
```diff
@@ -0,0 +1,72 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+function main(projectId, clientId, clientSecret, refreshToken) {
```
callmehiphop: Should we make fake default values?

samples/authUserFlow.js
```diff
@@ -0,0 +1,72 @@
...
+  const credentials = {
+    type: 'authorized_user',
+    client_id: clientId,
+    client_secret: clientSecret,
+    refresh_token: refreshToken,
+  };
+  // [END bigquery_auth_user_flow]
+  // [START bigquery_auth_user_query]
+  async function query() {
+    const {BigQuery} = require('@google-cloud/bigquery');
```
callmehiphop: Probably should move the `require` + `new` call outside of the query function.

samples/authUserFlow.js
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const projectId = 'PROJECT ID'
+  // const clientId = 'CLIENT ID'
+  // const clientSecret = 'CLIENT SECRET'
+  // const refreshToken = 'REFRESH TOKEN'
+
+  const credentials = {
+    type: 'authorized_user',
```
callmehiphop: I'm not sure if we actually support these options? I've never seen them documented anywhere (so I definitely could be wrong) but looking at the TypeScript interfaces it looks like the only options we accept here are `client_email` and `private_key`.

steffnay: I thought we didn't either! That's why I had this whole work-around when I first made the sample, but I tried it out based off a comment by @stephenplusplus and it appears to be working. I could be totally missing something, however, so please try it out and let me know, if you have time.

steffnay:
> LGTM, although I'm not sure the testing philosophy here. (cc @callmehiphop)
> 
> Do we need a new test for this? I know the query used in this sample is repeated throughout various tests. I'm not sure if the tests are repeated for each instance as well.
> 
> I assume the things we would want to test about this are:
> 
> 1. Is this type of authentication valid
> 2. Does the code that runs the query and returns the results work as expected
> 
> We can copy and paste the test for #­2, but #­1 is tricky. Would we have to intercept the auth library to make sure the `credentials` object was received, but return a normal auth instance that the other tests usually run with?

@stephenplusplus @callmehiphop  #1 is definitely tricky for me. I would need an example of the steps for intercepting the auth library because I don't think I've written tests that do something similar yet.

samples/authUserFlow.js
JustinBeckwith: I followed this all the way down the rabbit hole, and I actually think it would work.  The code in `googleauth.ts` is not good.  
https://github.com/googleapis/google-auth-library-nodejs/blob/master/src/auth/googleauth.ts#L448

If you follow that path, it should actually pull in the JSON you're passing here.

samples/authUserFlow.js
```diff
@@ -0,0 +1,72 @@
...
+
+function main(projectId, clientId, clientSecret, refreshToken) {
+  // [START bigquery_auth_user_flow]
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const projectId = 'PROJECT ID'
+  // const clientId = 'CLIENT ID'
+  // const clientSecret = 'CLIENT SECRET'
+  // const refreshToken = 'REFRESH TOKEN'
```
tswast: I want to be a little more careful about refresh tokens in particular. Client ID & secret only identify an app, so it's okay to embed those in code, but refresh tokens identify a user so it's not appropriate to embed in code.

Is there a way we can encourage people to pass these in via command-line or environment variable?

samples/authUserFlow.js
```diff
@@ -0,0 +1,72 @@
...
+    const query = `SELECT name
+            FROM \`bigquery-public-data.usa_names.usa_1910_2013\`
+            WHERE state = 'TX'
+            LIMIT 100`;
+
+    // For all options, see https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query
+    const options = {
+      query: query,
+      // Location must match that of the dataset(s) referenced in the query.
+      location: 'US',
```
tswast: We're removing all the location options from our query samples in Python because the location auto-detect works in every region now for queries.

samples/authUserFlow.js
tswast: I assume we'll have some other sample code somewhere for how to get a refresh token?

In the Python example, we have folks paste in an authentication code to standard-in, which we then exchange for a refresh token.

samples/authUserFlow.js
```diff
@@ -0,0 +1,189 @@
...
+
+// Download your OAuth2 configuration from the Google
+const keys = require('./oauth2.keys.json');
+
+/**
+ * Keep a config.keys.json with a username and password
+ * that correspond to your test google account:
+ * {
+ *   "username": "your-user-name@gmail.com",
+ *   "password": "your-password"
```
tswast: Is this necessary? I feel uncomfortable putting my password in such a file. Isn't it possible to post a link to `https://accounts.google.com/o/oauth2/auth?client_id=...` and then wait for user input on the authentication code?

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,156 @@
...
+    scope: 'https://www.googleapis.com/auth/bigquery',
+    prompt: 'consent',
+  });
+
+  console.info(
+    `Please visit this URL to authorize this application: ${authorizeUrl}`
+  );
+  rl.question('Enter the authorization code: ', function(code) {
+    main.exchangeCode(code).then(function() {
+      rl.close();
```
tswast: Once we exchange the code for a refresh token, we can use it below in "authFlow"

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,156 @@
...
+  };
+}
+// [END bigquery_auth_user_flow]
+// [START bigquery_auth_user_query]
+async function query() {
+  const {BigQuery} = require('@google-cloud/bigquery');
+  const credentials = main.authFlow();
+  const bigquery = new BigQuery(credentials);
+  // Queries the U.S. given names dataset for the state of Texas.
+  const query = ` SELECT name, SUM(number) as total
```
JustinBeckwith: Probably ok to use the single quotes here, so you don't need to escape the ticks.

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,156 @@
...
+    // Wait for the query to finish
+    const [rows] = await job.getQueryResults();
+
+    // Print the results
+    console.log('Rows:');
+    rows.forEach(row => console.log(row));
+
+    return rows;
+  } catch (err) {
+    console.log(err);
```
JustinBeckwith: No need for the try/catch here - just let it throw!

samples/auth-user-sample/test/auth.test.js
```diff
@@ -0,0 +1,99 @@
+// /**
```
JustinBeckwith: Double comments in this block

samples/auth-user-sample/test/auth.test.js
```diff
@@ -0,0 +1,99 @@
...
+//  * Unless required by applicable law or agreed to in writing, software
+//  * distributed under the License is distributed on an "AS IS" BASIS,
+//  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  * See the License for the specific language governing permissions and
+//  * limitations under the License.
+//  */
+
+// 'use strict';
+
+const mocha = require('mocha');
```
JustinBeckwith: Generally don't have to explicitly require mocha like this, or assign these below.  They're just ambiently available

samples/auth-user-sample/test/auth.test.js
```diff
@@ -0,0 +1,99 @@
...
+  after(async () => {
+    await bigquery
+      .dataset(datasetId)
+      .delete({force: true})
+      .catch(console.warn);
+  });
+
+  it('should exchange code for tokens', async function() {
+    const output = await authUserFlow.main.exchangeCode('abc123');
+    sinon.assert.match(output, 'tokens');
```
JustinBeckwith: I'm a little confused about the usage of `sinon.assert` here.  How is this different from a regular `assert`?

samples/auth-user-sample/test/auth.test.js
```diff
@@ -0,0 +1,99 @@
...
+    });
+  });
+  after(async () => {
+    await bigquery
+      .dataset(datasetId)
+      .delete({force: true})
+      .catch(console.warn);
+  });
+
+  it('should exchange code for tokens', async function() {
```
JustinBeckwith: No need to say `async function() {` here - instead, do `async () => {`

samples/auth-user-sample/test/auth.test.js
steffnay: The docs for testing whether stubs were called use sinon.assert, so I went with that. I haven't seen how to do the same with regular assert.

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,156 @@
...
+  const authorizeUrl = oAuth2Client.generateAuthUrl({
+    access_type: 'offline',
+    scope: 'https://www.googleapis.com/auth/bigquery',
+    prompt: 'consent',
+  });
+
+  console.info(
+    `Please visit this URL to authorize this application: ${authorizeUrl}`
+  );
+  rl.question('Enter the authorization code: ', function(code) {
```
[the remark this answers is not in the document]
steffnay: I switched to the readline-promise library so I can use async/await

samples/auth-user-sample/test/auth.test.js
steffnay: My linter says no-undefined when I don't do these assignments. Is there something else I should be doing about that?

samples/auth-user-sample/authUserFlow.js
steffnay: I think it needs to be these ticks because it's consistent with the rest of the queries in the node samples.

samples/auth-user-sample/authUserFlow.js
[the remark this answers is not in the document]
steffnay: Just kidding, I switched from using this altogether

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,156 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const {OAuth2Client} = require('google-auth-library');
+const readline = require('readline');
```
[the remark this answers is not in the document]
steffnay: Gotcha. Thanks!

samples/auth-user-sample/test/auth.test.js
bcoe: I think that this is equivalent to:

```js
assert.strictEqual(output, 'tokens');
```

Took me for a slight loop too, just because I was trying to figure out whether sinon was doing something fancy, e.g., spying on `output` or something. I'd probably opt for `assert` like @JustinBeckwith suggests, because it communicates the fact that this is just a simple assertion.

steffnay: Yeah, this is correct, I thought this was asking about using sinon.assert when checking to see if a stub was called. I misread. Should I be using assert for these simple comparisons and sinon.assert for checking stubs? Or is there another way to use assert to just verify that stubbed out portions of the authorization flow are being called?

samples/auth-user-sample/authUserFlow.js
steffnay: This was helpful and I refactored to do it this way.... Then I got other feedback and changed it again....

samples/auth-user-sample/test/auth.test.js
bcoe: As a general rule, I'd use Sinon's methods for checking if methods are called, which it's great at, e.g.,

`sinon.assert.called(authenticateStub);` which seems like great use of sinon's assertions. For simple comparisons, I'd probably opt for `assert()`, `assert.strictEqual()`, etc.

steffnay: Gotcha. Thanks.

samples/auth-user-sample/authUserFlow.js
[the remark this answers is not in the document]
tswast: If we had to pick one, I actually prefer the readline version. That way it works in every environment, including Cloud Shell.

In Python, we provide both, but it's a lot easier because it's built into the auth libs.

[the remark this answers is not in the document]
steffnay: I do prefer writing the sample to work in all environments. I think I'll go back to the readline version just so we can get something published, then I can see if it makes sense to have both examples

samples/auth-user-sample/test/auth.test.js
JustinBeckwith: Yes! You can modify `samples/auth-user-sample/test/.eslintrc.yml to include:

```
---
env:
  mocha: true
```

This signals to `eslint` that you're using the ambient functions available in mocha, and will leave you alone :)

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,62 @@
+/**
```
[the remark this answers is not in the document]
steffnay: Will do!

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,156 @@
...
+    keys.installed.client_id,
+    keys.installed.client_secret,
+    keys.installed.redirect_uris[0]
+  );
+
+  try {
+    const r = await oAuth2Client.getToken(code);
+    console.info(r.tokens);
+    return r.tokens;
+  } catch (err) {
```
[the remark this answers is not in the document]
steffnay: Totally makes sense. Will change.

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,62 @@
...
+  return rows;
+}
+
+async function run() {
+  await sampleClient.authenticate(['https://www.googleapis.com/auth/bigquery']);
+  const credentials = sampleClient.oAuth2Client.credentials;
+  return query(credentials);
+}
+
+// run();
```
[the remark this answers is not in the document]
steffnay: Thank you!

samples/auth-user-sample/test/auth.test.js
steffnay: Thanks!

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,151 @@
...
+function startRl() {
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  return rl;
+}
+
+// Download your OAuth2 configuration from the Google
```
tswast: Missing some text here.

samples/.eslintrc.yml
```diff
@@ -1,3 +1,4 @@
 ---
 rules:
   no-console: off
+  no-missing-require: off
```
JustinBeckwith: We shouldn't need this

samples/auth-user-sample/.eslintrc.yml
```diff
@@ -0,0 +1,3 @@
+---
+env:
```
JustinBeckwith: This should be moved to the `test/` directory.  We don't want ambient mocha allowed in the directory with the source itself

samples/auth-user-sample/authUserFlow.js
JustinBeckwith: Generally you only use the backtick when you want to do string interpolation, like `hello ${name}`.  Otherwise, it's generally customary to use single quotes.

samples/auth-user-sample/authUserFlow.js
```diff
@@ -0,0 +1,155 @@
...
+  WHERE name = 'William'
+  GROUP BY name;`;
+
+  // For all options, see https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query
+  const options = {
+    query: query,
+  };
+
+  // Run the query as a job
+  try {
```
JustinBeckwith: No need to for the try catch here.  Just let it throw!

samples/auth-user-sample/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "test": "mocha --timeout 200000"
+  },
+  "dependencies": {
+    "@google-cloud/bigquery": "^4.4.0",
+    "google-auth-library": "^5.5.1",
+    "readline-promise": "^1.0.4"
+  },
+  "devDependencies": {
+    "chai": "^4.2.0",
+    "sinon-chai": "^3.3.0",
```
JustinBeckwith: don't think we need this

samples/auth-user-sample/test/auth.test.js
```diff
@@ -0,0 +1,81 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const sinon = require('sinon');
+const {assert} = require('chai');
+const proxyquire = require('proxyquire');
+
+describe('authUserFlow()', function() {
```
JustinBeckwith: Let's use an arrow function

samples/auth-user-sample/test/auth.test.js
```diff
@@ -0,0 +1,81 @@
...
+          return {
+            generateAuthUrl: generateUrlStub,
+            getToken: tokenStub,
+          };
+        }),
+      },
+    });
+  });
+
+  afterEach(function() {
```
JustinBeckwith: arrow function (here and below)

samples/.eslintrc.yml
steffnay: Because this is a sample app, the package.json is nested and wasn't getting installed properly by the CI so Ben Coe suggested it for this case.

samples/auth-user-sample/.eslintrc.yml
steffnay: Done

samples/auth-user-sample/authUserFlow.js
steffnay: I think this library uses the backticks specifically for the SQL queries. It doesn't work if you use single or double quotes instead.

samples/auth-user-sample/package.json
```diff
@@ -0,0 +1,29 @@
+{
```
JustinBeckwith: In most of our other sample directories, we tend to have a single `samples/package.json` that holds the dependencies you need for all the samples.  I think having this bonus `package.json` is what's causing some of the CI weirdness.  Thinking maybe we should:
- delete this file
- move any dependencies you need to `samples/package.json`

Guessing that will help clean a fair bit of this up. @bcoe has the final call though :)

steffnay: I was told that a sample app should have its own package.json, but I don't know if there's actually a standard for this!

bcoe: If the goal is to demonstrate to the user a fully self contained application, I don't think it's unreasonable to have the inner package.json (this reminds me of how we've approached samples in our old samples mono repo):

https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/appengine/cloudsql

If the goal is that we'd be writing upstream tutorials that embed this package.json, I think it's okay to have it in isolation. If this is not a requirement, it might be worth collapsing the dependencies into the `samples/package.json`.

tswast: I think the most important thing is to show the install steps. Probably doesn’t need to embed the package.json if we add a “clone the repo” step to the tutorial.

samples/.eslintrc.yml
steffnay: Got it! Done

samples/auth-user-sample/package.json
steffnay: Alrighty, sounds like it's a good idea to just move dependencies to samples/package.json.

steffnay:
@JustinBeckwith @bcoe I'm unsure who to check with regarding this. I'm curious whether the lint rules changed somehow in CI. When I lint and run npm fix locally, it styles one of my code samples in a way that fails the linter in my PR. Wondering how to sync the two up.

bcoe:
@steffnay there's a chance your local installation of `eslint` is out of sync with what is running in CI/CD:

```
find . -name "node_modules" -exec rm -rf '{}' +
find . -name "package-lock.json" -exec rm -rf '{}' +
npm i
npm run lint
npm run fix
```

That should do the trick for you, let me know if you bump into any issues.

samples/auth-user-sample/authUserFlow.js
steffnay: We've been taking this approach with our SQL samples, because it allows us to spread them
across multiple lines; would you be open to us sticking to this approach for consistency?

JustinBeckwith: It's fine 🤷‍♂

Issue #568, opened 2019-11-06
Add option to use environment variable to set baseUrl for BigQuery constructor


amalhotragannett:
Problem: In order to make testing more efficient we need to create a mock BigQuery server to simulate errors and success. Currently, we creating the BigQuery constructor we can pass in `apiEndpoint` to override the URL. However, when the baseUrl is created it uses the `https` protocol. We want to override the value using environment variables. 

Alternatives Considered: 
`apiEndpoint` to override the endpoint.
Example:
```
const bq = new BigQuery({
      apiEndpoint: 'localhost:9090'
    });
```
This makes sets the `bq.baseUrl='https://localhost:9090'`
Since, the local emulator is running on `http://localhost:9090`. The request are not received due to SSL. 

The emulator is not provided by Google. We are using a custom application which response to a request from BigQuery APIs. 

Expected Behavior:
A configuration through environment variable or an option to override baseUrl.

callmehiphop:
Thanks for opening an issue! I was looking at some of our other client libraries to see if/how we supported specifying protocols elsewhere and it looks like in some cases we offer a `protocol` option.

@alexander-fenster @bcoe what are your feelings on adding support for `protocol` here and allowing users to configure it? I'm not sure if that is the intended use for it elsewhere, since I could only find references to it for grpc fallback code.

amalhotragannett:
@callmehiphop - I was looking into the Google Cloud Storage for the emulator HOST. 
https://github.com/googleapis/nodejs-storage/blob/862fb16db2990c958c0097252a44c775431a7b3f/src/storage.ts#L375-L390

It didn't have protocol support. But it did have a similar environment variable to incorporate the protocols.

callmehiphop:
Interesting, @tswast @shollyman do either of you have any feelings on supporting an environment variable for third party emulators?

stephenplusplus:
Note that there's some oddity in how that is implemented in Storage: https://github.com/googleapis/nodejs-storage/issues/898#issuecomment-547461664

amalhotragannett:
There is a similar implementation with the [PubSub package](https://github.com/googleapis/nodejs-pubsub/blob/1749b626e6bff5fefd1b1b8c673c480a10be9cf9/src/pubsub.ts#L524-L546) as well.

shollyman:
[...]

PR #569, opened 2019-11-06
add option to override baseURL using environment variables

amalhotragannett:
- [ ] Ensure the tests and linter pass

Fixes [#568](https://github.com/googleapis/nodejs-bigquery/issues/568)🦕

### DIFF
--- a/src/bigquery.ts
+++ b/src/bigquery.ts
@@ -245,9 +245,12 @@ export class BigQuery extends common.Service {
 
   constructor(options: BigQueryOptions = {}) {
     options.apiEndpoint = options.apiEndpoint || 'bigquery.googleapis.com';
+    const url =
+      process.env.BIGQUERY_EMULATOR_HOST ||
+      `https://${options.apiEndpoint}/bigquery/v2`;
     const config = {
       apiEndpoint: options.apiEndpoint,
-      baseUrl: `https://${options.apiEndpoint}/bigquery/v2`,
+      baseUrl: url,
       scopes: ['https://www.googleapis.com/auth/bigquery'],
       packageJson: require('../../package.json'),
     };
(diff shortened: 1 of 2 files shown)


src/bigquery.ts
```diff
@@ -245,9 +245,12 @@ export class BigQuery extends common.Service {
 
   constructor(options: BigQueryOptions = {}) {
     options.apiEndpoint = options.apiEndpoint || 'bigquery.googleapis.com';
+    const url =
```
JustinBeckwith: @tswast @shollyman  this feels like something we should do consistently across clients, if we want to do it.  What do we do in other languages?

tswast: This was already completed in Python via the "client options" property. go/extensible-options-in-client

tswast: Hmm... Not seeing a design for Node.js in that internal doc, though. Just Go & Python, but it definitely needs to be done across all languages.

amalhotragannett: @JustinBeckwith would this require a design change for the library. We are about to migrate our CI pipeline to cloud build and were trying to avoid the route of using the SAN certs for our integration tests.

bcoe: This does come up occasionally for all of our emulators. I wonder if we could introduce into the auth library, `EMULATOR_HOST`.

bcoe: This does come up occasionally for all of our emulators. I wonder if we could introduce into the auth library, `EMULATOR_HOST`.

bcoe: This does come up occasionally for all of our emulators. I wonder if we could introduce into the auth library, `EMULATOR_HOST`.

JustinBeckwith:
@steffnay @shollyman could I trouble y'all to take a peek here?  What do we want to do with this?

PR #582, opened 2019-11-21
feat(table): allow opting out of default insert id

### LINKED ISSUE #581: BigQuery: provide a way to opt-out of sending row IDs when streaming rows to a BigQuery table
To get 1,000,000 rows per second quota, the row ID needs to be omitted. See: https://cloud.google.com/bigquery/quotas#streaming_inserts

Continue to provide row IDs by default, as that reduces the number of duplicate rows. Allow the developer to opt-in to the higher quotas by explicitly asking for no row ID. In Python we can check if None is supplied as the row ID and use that. In Go, that's not possible so they use `"NoDedupeID"` as an indicator for this behavior.

There's an argument to be made that row IDs could be omitted by default, but this could result in many more duplicate rows for the majority of developers who do not require the higher quotas.

 **Additional context**

* C# issue: https://github.com/googleapis/google-cloud-dotnet/issues/3769
* Go commit: https://github.com/googleapis/google-cloud-go/commit/1366340a586e4da4c221c5c7dee6f08c977864dc
* Python issue: https://github.com/googleapis/google-cloud-python/issues/9539

callmehiphop:

Fixes #581 🦕

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -1885,6 +1885,7 @@ class Table extends common.ServiceObject {
       });
     }
 
+    delete json.createInsertId;
     delete json.raw;
 
     let schema: string | {};
--- a/src/table.ts
+++ b/src/table.ts
@@ -69,6 +69,7 @@ export type JobMetadataResponse = [JobMetadata];
 export type RowMetadata = any;
 
 export type InsertRowsOptions = bigquery.ITableDataInsertAllRequest & {
+  createInsertId?: boolean;
   raw?: boolean;
   schema?: string | {};
 };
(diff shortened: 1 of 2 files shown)


callmehiphop:
@tswast hopefully I understood the request!

tswast (approved):
Great! Thanks.

src/table.ts
```diff
@@ -1861,16 +1869,19 @@ class Table extends common.ServiceObject {
...
     if (!options.raw) {
       json.rows = rows.map((row: RowMetadata) => {
-        return {
-          insertId: uuid.v4(),
-          json: Table.encodeValue_(row),
+        const encoded: InsertRow = {
+          json: Table.encodeValue_(row)!,
         };
+
+        if (options.createInsertId !== false) {
```
stephenplusplus: A few lines below, “json.raw” is deleted. Maybe this option should be removed from the object that is sent upstream as well?

callmehiphop: Good call! Just added a commit for that

PR #589, opened 2019-12-09
fix(table): add retries for insert partial failures

callmehiphop:


src/table.ts
```diff
@@ -133,6 +133,18 @@ export type FormattedMetadata = bigquery.ITable;
 export type TableSchema = bigquery.ITableSchema;
 export type TableField = bigquery.ITableFieldSchema;
 
+interface PartialFailureError extends Error {
```
callmehiphop: Ignore this, it needs to be removed.

bcoe:
@callmehiphop is there an issue or internal document related to this, curious to have a bitt of background.

callmehiphop:
@bcoe there is an internal issue, but I couldn't tell you where to find it. @steffnay could probably help here

meredithslota:
b/145598828 is the internal issue.

shollyman (approved):
This seems like a reasonable approach (finite retry when there are entries in the structured error response from invoking tabledata.insertAll).

src/table.ts
```diff
@@ -1886,74 +1984,39 @@ class Table extends common.ServiceObject {
...
-      (err, resp) => {
-        if (err) {
-          if ((err as common.ApiError).code === 404 && schema) {
-            setTimeout(createTableAndRetry, Math.random() * 60000);
-          } else {
-            callback!(err, resp);
-          }
-          return;
-        }
+    const [resp] = await this.request({
```
zamnuts: @steffnay it appears that this return from `request` was refactored by @callmehiphop, where request is now returning an array (tuple) containing the response, but previously this was not the case (as you can see by this diff). It appears this was a good fix per `@google-cloud/common`'s `service-object.d.ts`:
```
export declare type RequestResponse = [Metadata, r.Response];
```
But I'm having trouble confirming if this is really the case. Can you verify?

zamnuts:
@googlebot I consent

src/table.ts
```diff
@@ -1863,6 +1872,95 @@ class Table extends common.ServiceObject {
...
+    try {
+      await this.create({schema});
+    } catch (e) {
+      if (e.code !== 409) {
+        throw e;
+      }
+    }
+
+    await new Promise(resolve => setTimeout(resolve, delay));
+    return (await this.insert(rows, options))[0];
```
JustinBeckwith: Could we make this a little more readable?  This could be 2-3 lines of code that more sense :)

src/table.ts
```diff
@@ -1863,6 +1872,95 @@ class Table extends common.ServiceObject {
...
+  }
+
+  /**
+   * This method will attempt to insert rows while retrying any partial failures
+   * that occur along the way. Because partial insert failures are returned
+   * differently, we can't depend on our usual retry strategy.
+   *
+   * @private
+   *
+   * @param {object|object[]} rows The rows to insert.
```
JustinBeckwith: The jsdoc here is not correct.  This should refer to a `RowMetadata` object that needs to be defined both as an interface, and as a jsdoc annotation.  I know this is terrible.

test/table.ts
```diff
@@ -1178,9 +1187,31 @@ describe('BigQuery/Table', () => {
       });
     });
 
+    // TODO this doesn't work without promises or callbacks, refactor as
```
JustinBeckwith: We're going to want to fix this before we land :)

test/table.ts
```diff
@@ -2014,54 +2045,104 @@ describe('BigQuery/Table', () => {
       }),
     };
 
+    // HACK @types/sinon is missing the timer async methods
```
JustinBeckwith: So it's ok to have this here for now.  Please:
- submit an upstream PR to Definitely Typed with the fix
- reference the PR in the comment here
- file a process bug to follow up after the fix lands

We try to help maintain the community types here :)

test/table.ts
```diff
@@ -2014,54 +2045,104 @@ describe('BigQuery/Table', () => {
...
+     * when relying on any of the fake timer async helpers.
+     * Tests should assert isRejected or isFulfilled.
+     * @ignore
+     * @param fn
+     * @returns {Promise<pReflect.PromiseResult<any>>}
+     */
+    async function reflectAfterTimer<FnReturn>(
+      fn: () => Promise<FnReturn>
+    ): Promise<pReflect.PromiseResult<FnReturn>> {
+      /*
```
JustinBeckwith: So this is a new one :) Could I trouble you to use a more traditional comment style?  Either:

```js
/**
 * I am a multiline comment rawr
 */
```

Or ...

```js
// this
// is
// fine
// too
```

JustinBeckwith (dismissed):

src/table.ts
steffnay: @zamnuts Yeah, it looks like that's correct.

test/table.ts
zamnuts: This is related to #640, i'll bundle it with this PR then.

test/table.ts
zamnuts: i filed https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43148

and upon submitting the PR realized https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43139 was also open, less than 12h difference :(

zamnuts: Looks like it's going to be a while until the DT PR is merged, due to an export bug in a dependency (`@sinonjs/fake-timers` proper). I've opened #655 to track it.

zamnuts:
Waiting on changes from #647 to resolve codecov decrease and to remove test TODO

src/table.ts
zamnuts: refactored as requested.

src/table.ts
zamnuts: updated affected jsdocs

test/table.ts
zamnuts: #640 is resolved via #647 and has been merged to master. this test is no longer valid and has been removed

test/table.ts
zamnuts: yeah, that's my bad, changed to `//`

test/table.ts
```diff
@@ -2080,215 +2128,198 @@ describe('BigQuery/Table', () => {
...
-      const apiResponse = {};
-
-      table.request = (reqOpts: DecorateRequestOptions, callback: Function) => {
-        callback(error, apiResponse);
-      };
+      requestStub.rejects(error);
 
       table.insert(data, (err: Error, apiResponse_: {}) => {
         assert.strictEqual(err, error);
-        assert.strictEqual(apiResponse_, apiResponse);
```
stephenplusplus: Do we lose getting back the "apiResponse" argument when there's an error? Previously, we returned the "err" and "resp" from both `this.create()` and `this.request()` when they returned an error.

src/table.ts
```diff
@@ -1872,6 +1887,98 @@ class Table extends common.ServiceObject {
...
+   *
+   * @param {RowMetadata|RowMetadata[]} rows The rows to insert.
+   * @param {InsertRowsOptions} options Insert options.
+   * @returns {Promise<bigquery.ITableDataInsertAllResponse>}
+   */
+  private async _insertWithRetry(
+    rows: RowMetadata | RowMetadata[],
+    options: InsertRowsOptions
+  ): Promise<bigquery.ITableDataInsertAllResponse> {
+    const maxAttempts = options.maxAttempts || 3;
```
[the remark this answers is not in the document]
JustinBeckwith: Nice catch!

test/table.ts
```diff
@@ -2021,55 +2022,102 @@ describe('BigQuery/Table', () => {
...
-
-      table.insert([data[0]], {createInsertId: false}, done);
+    it('should omit the insertId if createInsertId is false', async () => {
+      await table.insert([data[0]], {createInsertId: false});
+      assert(requestStub.calledOnce);
+      assert(
+        requestStub.calledWithMatch(
+          ({json}: DecorateRequestOptions) =>
+            json.rows[0].insertId === undefined &&
+            json.createdInsertId === undefined
```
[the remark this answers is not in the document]
zamnuts: oh my, thank you, good catch. no typing for `DecorateRequestOptions`'s `json` property (`json: any`), and see what happens!

src/table.ts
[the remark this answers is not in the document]
zamnuts: currently `maxAttempts` includes the _first_ request, regardless of failure; so if a user wanted to try once, then retry up to 3 more times, `maxAttempts` would have to be 4. A value of `0` means don't even try, which is probably not what we want.

Additionally, this retry is specifically for _partial row insertion failures_ and not any other failure scenarios.

I've changed the option to `partialRetries`, modeled off of [`node-retry`](https://github.com/tim-kos/node-retry#retryoperationoptions and scoped to the "partial" concept, to avoid confusion.

test/table.ts
zamnuts: Yes, promise rejections only allow a single error object. The api response could be added as a property of the error object, but that is not a pattern used in these node repos. This was due to the additional abstractions provided early-on in this PR for asserting the table, and handling the retries.

test/table.ts
zamnuts: `@types/sinon@9.0.0` landed, merged into this PR, and I was able to remove the shim.

stephenplusplus (approved):

steffnay (approved):

PR #599, opened 2019-12-26
feat(params): adds optional param types

### LINKED ISSUE #400: Named query params with an empty array cannot be parsed
#### Environment details

  - OS: High Sierra (10.13.6)
  - Node.js version: 8.15.1
  - npm version: 6.9.0
  - `@google-cloud/bigquery` version: 2.1.0

#### Steps to reproduce

  1. Run a query with an empty array as the value of a named query param, e.g.
```ts
const [rows] = await bq.query({
  query,
  params: {
    key: [],
  },
});
```
  2. Query fails with `This value could not be translated to a BigQuery data type`.
  3. Attempt to run the same query with an empty array literal in BigQuery UI or CLI and observe that the query succeeds:
```sql
...
WHERE `dataset.table`.Key IN UNNEST([])
```

#### Source of the error
I believe the error is a result of the login in https://github.com/googleapis/nodejs-bigquery/blob/ba5c82efc4c6f2aa50df2c386ab0b6894d0dd2aa/src/index.ts#L726

You can see that `nodejs-bigquery` is trying to determine the type of the query param from the first element. If the array is empty, `value[0]` is `undefined`, and `getType_()` throws an error. This issue is similar to https://github.com/googleapis/google-cloud-java/issues/2678, where they did fix it (and as they state there, it is not a limitation of the API).

EDIT: The API does seem to require specifying `arrayType`, e.g.
```json
"parameterType": {
  "type": "ARRAY",
  "arrayType": {
    "type": "INT64"
  }
},
```
Since JavaScript (unlike Java) doesn't store the array type info at runtime (afaik), I wonder if there's a way to supply the type explicitly in this case. Not sure if this is more of a feature request and would you consider a PR for it.

Thanks!

steffnay:
Fixes #400 🦕


src/bigquery.ts
```diff
@@ -729,6 +732,60 @@ export class BigQuery extends common.Service {
...
+   * @param {*} providedType The type.
+   * @returns {string} The valid type provided.
+   */
+  // tslint:disable-next-line no-any
+  static getProvidedType_(providedType: any): ValueType {
+    const VALID_TYPES = [
+      'DATE',
+      'DATETIME',
+      'TIME',
+      'TIMESTAMP',
```
bcoe: not a blocker, but I wonder if there's any way we could use the `IStandardSqlDataType` from `./src/types.d.ts` for this list?

We might also consider this as the type for `providedType` rather than `any`.

alexander-fenster:
I don't know much about this library (it's not a GAPIC), but I generally agree that we should avoid using `any`. It might make sense to have this one merged and then have a separate PR to get rid of all explicitly used `any` types in the code since I'm guessing most of the types should present in `.d.ts`.

src/bigquery.ts
```diff
@@ -758,10 +819,18 @@ export class BigQuery extends common.Service {
     } else if (value instanceof Big) {
       typeName = 'NUMERIC';
     } else if (is.array(value)) {
-      return {
-        type: 'ARRAY',
-        arrayType: BigQuery.getType_(value[0]),
-      };
+      if (value.length === 0 || value[0] === null) {
```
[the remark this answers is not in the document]
steffnay: As far as I know, undefined shouldn't be an option, just null

src/bigquery.ts
steffnay: Could you possibly give me an example of what that would look like? I still have some issues with with types in TypeScript

steffnay: Also, I'm curious of you could take a peek at the tests. I feel like this isn't actually working like I want it to

bcoe: Digging into it, there might not be an easy way to do this, what I might just do is add a comment that says the list of types can be found in `/src/types.d.ts`, my only other question, do we need to allow the other enums listed:

```
      | 'NUMERIC'
      | 'ARRAY'
      | 'STRUCT';
```?

Comment (steffnay) [reply src/bigquery.ts]: Okay, got it.

File: src/bigquery.ts [hunk 3]
```diff
@@ -729,6 +732,60 @@ export class BigQuery extends common.Service {
...
+            type: BigQuery.getProvidedType_(providedType[prop]),
+          };
+        }),
+      };
+    } else if (is.boolean(providedType)) {
+      return {type: providedType};
+    }
+
+    if (!VALID_TYPES.includes(providedType.toUpperCase())) {
+      throw new Error('Invalid type provided.');
```
stephenplusplus: Could you put the value the user provided here, e.g. `Invalid type provided: "INT63"`

src/bigquery.ts
```diff
@@ -729,6 +732,60 @@ export class BigQuery extends common.Service {
...
+      return {
+        type: 'STRUCT',
+        structTypes: Object.keys(providedType).map(prop => {
+          return {
+            name: prop,
+            type: BigQuery.getProvidedType_(providedType[prop]),
+          };
+        }),
+      };
+    } else if (is.boolean(providedType)) {
```
stephenplusplus: Is it valid/expected for a user to provide a boolean as a type?

src/bigquery.ts
```diff
@@ -742,9 +799,13 @@ export class BigQuery extends common.Service {
    * @returns {string} The type detected from the value.
    */
   // tslint:disable-next-line no-any
-  static getType_(value: any): ValueType {
+  static getType_(value: any, providedType?: any): ValueType {
     let typeName;
 
+    if (providedType) {
+      return this.getProvidedType_(providedType);
```
stephenplusplus: The naming here confused me a bit. Just for consideration, what about:

```diff
- getType_
+ getTypeDescriptorFromValue_

- getProvidedType_
+ getTypeDescriptorFromProvidedType_
```

Definitely longer, but I think something like that would help someone jump in for a bug fix more quickly.

src/bigquery.ts
```diff
@@ -805,12 +874,12 @@ export class BigQuery extends common.Service {
...
    */
   // tslint:disable-next-line no-any
-  static valueToQueryParameter_(value: any) {
+  static valueToQueryParameter_(value: any, providedType: any = undefined) {
     if (is.date(value)) {
       value = BigQuery.timestamp(value as Date);
     }
 
-    const parameterType = BigQuery.getType_(value);
+    const parameterType = BigQuery.getType_(value, providedType);
```
stephenplusplus: Could we do an if/else for calling the "getProvidedType_" path when `providedType` is defined, and the "getType_" when it isn't?

src/bigquery.ts
```diff
@@ -1053,18 +1127,52 @@ export class BigQuery extends common.Service {
...
+
+          if (query.types) {
+            if (!is.object(query.types)) {
+              throw new Error(
+                'Provided types must match the value type passed to `params`'
+              );
+            }
+            queryParameter = BigQuery.valueToQueryParameter_(
+              value,
+              query.types[namedParameter]
```
stephenplusplus: Would it be throwable if this returned undefined, e.g. a user provided the types object, but only had types for some (or none) of the parameters? (Same question for the array version of code beneath this section)

stephenplusplus:
@steffnay sorry for not focusing on the testing right away, I wanted to make sure I understood the PR to begin with. Thanks for your patience with my questions!

src/bigquery.ts
```diff
@@ -42,6 +42,7 @@ import {
 } from './table';
 import {GoogleErrorBody} from '@google-cloud/common/build/src/util';
 import bigquery from './types';
+import {type} from 'os';
```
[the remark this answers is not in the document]
steffnay: Nope

src/bigquery.ts
```diff
@@ -94,6 +95,8 @@ export type Query = JobRequest<bigquery.IJobConfigurationQuery> & {
   // tslint:disable-next-line no-any
   params?: any[] | {[param: string]: any};
   dryRun?: boolean;
+  // tslint:disable-next-line: no-any
+  types?: any[] | {[type: string]: any};
```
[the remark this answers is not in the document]
steffnay: You're right, I can change this.

src/bigquery.ts
steffnay: Yeah, it's a BigQuery type that can be detected so they should be able to provide it as well.

src/bigquery.ts
steffnay: Good idea. Done.

src/bigquery.ts
steffnay: Done.

src/bigquery.ts
steffnay: Done.

src/bigquery.ts
steffnay: I'll add this.

steffnay:
@stephenplusplus thanks for the review. Just let me know if you spot any other changes to be made before you take a look at the tests!

test/bigquery.ts
```diff
@@ -1291,6 +1323,53 @@ describe('BigQuery', () => {
...
+            {
+              query: QUERY_STRING,
+              params: NAMED_PARAMS,
+              types: NAMED_TYPES,
+            },
+            assert.ifError
+          );
+        });
+
+        it('should throw for invalid provided types', done => {
```
stephenplusplus: This test should only be in the describe block for getTypeDescriptorFromProvidedType_.

test/bigquery.ts
```diff
@@ -1291,6 +1323,53 @@ describe('BigQuery', () => {
...
+            bq.createQueryJob({
+              query: QUERY_STRING,
+              params: NAMED_PARAMS,
+              types: INVALID_TYPES,
+            });
+          }, /Invalid type provided./);
+        });
+
+        it('should throw for invalid type structure provided', done => {
+          bq.createJob = (reqOpts: JobOptions) => {
```
stephenplusplus: I would remove this override since this hopefully won't get called. Another option would be executing done with an error, `done(new Error('Should not have been called.'))`.

test/bigquery.ts
```diff
@@ -1332,6 +1411,79 @@ describe('BigQuery', () => {
...
+            {
+              query: QUERY_STRING,
+              params: POSITIONAL_PARAMS,
+              types: POSITIONAL_TYPES,
+            },
+            assert.ifError
+          );
+        });
+
+        it('should throw for invalid provided types', done => {
```
stephenplusplus: This test should only be in the describe block for getTypeDescriptorFromProvidedType_.

test/bigquery.ts
```diff
@@ -1332,6 +1411,79 @@ describe('BigQuery', () => {
...
+          assert.throws(() => {
+            bq.createQueryJob({
+              query: QUERY_STRING,
+              params: POSITIONAL_PARAMS,
+              types: NAMED_TYPES,
+            });
+          }, /Provided types must match the value type passed to `params`/);
+        });
+
+        it('should throw for empty array without provided types', done => {
```
stephenplusplus: This test should only be in the describe block for getTypeDescriptorFromProvidedType_.

test/bigquery.ts
```diff
@@ -1332,6 +1411,79 @@ describe('BigQuery', () => {
...
+
+          assert.throws(() => {
+            bq.createQueryJob({
+              query: QUERY_STRING,
+              params: [],
+            });
+          }, /Type must be provided for empty array./);
+        });
+
+        it('should throw for null value without provided types', done => {
```
stephenplusplus: This test should only be in the describe block for getTypeDescriptorFromProvidedType_.

stephenplusplus:
Sorry for the delay @steffnay.

I believe we'll want a new `describe` block just to test `getTypeDescriptorFromProvidedType_`. That would handle all of the checks for if it throws when expected, recurses itself as intended, etc. Tests for other describes can then remain isolated to only verify behavior that happens in their respective functions.

I wrote up a working example of how you can verify that a sibling method is being called with the correct arguments, and that the value it returns is being used properly. This template should be repeatable for the different code paths, but let me know if you get stuck or have more questions about how to apply this elsewhere.

This is a test that would go under createQueryJob -> SQL parameters -> positional.

```ts
it('should convert value and type to query parameter', done => {
  const fakeQueryParameter = {fake: 'query parameter'};

  bq.createJob = (reqOpts: JobOptions) => {
    const queryParameters = reqOpts.configuration!.query!.queryParameters;
    assert.deepStrictEqual(queryParameters, [fakeQueryParameter]);
    done();
  };

  sandbox
    .stub(BigQuery, 'valueToQueryParameter_')
    .callsFake((value, type) => {
      assert.strictEqual(value, POSITIONAL_PARAMS[0]);
      assert.strictEqual(type, POSITIONAL_TYPES[0]);
      return fakeQueryParameter;
    });

  bq.createQueryJob({
    query: QUERY_STRING,
    params: POSITIONAL_PARAMS,
    types: POSITIONAL_TYPES,
  });
});
```

I hope this is helpful!

src/bigquery.ts
steffnay: Oh, I see. Fixed.

test/bigquery.ts
```diff
@@ -708,26 +735,207 @@ describe('BigQuery', () => {
...
+      const INVALID_TYPES = {key: 'invalid'};
+
+      assert.throws(() => {
+        bq.createQueryJob({
+          query: QUERY_STRING,
+          params: NAMED_PARAMS,
+          types: INVALID_TYPES,
+        });
+      }, /Invalid type provided./);
+      done();
```
stephenplusplus: This done() and the ones that follow in this block can be removed. It's only needed when we want to hold the test from exiting prematurely while we wait for a callback to be executed. In this case, the `assert.throws()` will run sync and then exit the test block with a failure or a success.

test/bigquery.ts
```diff
@@ -708,26 +735,207 @@ describe('BigQuery', () => {
...
+      assert.throws(() => {
+        bq.createQueryJob({
+          query: QUERY_STRING,
+          params: null,
+        });
+      }, /Type must be provided for null values./);
+      done();
+    });
+
+    it('should return correct types', () => {});
```
stephenplusplus: TODO?

test/bigquery.ts
```diff
@@ -708,26 +735,207 @@ describe('BigQuery', () => {
...
+        });
+      }, /Invalid type provided./);
+      done();
+    });
+
+    it('should throw if named param not present in provided types', done => {
+      const INVALID_TYPES = {other: 'string'};
+
+      assert.throws(() => {
+        bq.createQueryJob({
```
stephenplusplus: Sorry if I was confusing with my earlier review. Essentially, the goal is to separate tested methods as much as possible, so if we change behavior in `getTypeDescriptorFromProvidedType_()` one day, we don't have to update the tests for `createQueryJob()`. So, the tests in this describe block should call `getTypeDescriptorFromProvidedType_` directly.

When it comes to testing methods that call each other, we rely on the mocked version of the called method to verify that it is called as we expect and that the value it returns is used as we intend. That way, we don't have to know anything about what the called method is doing. It has its own `describe` that will make sure it works properly.

And then when it comes to making sure the "big picture" behavior works as intended, we have our system tests that make sure the whole machine is working as we advertise.

The tests listed below are testing functionality that lives in `createQueryJob()`, so they should be moved over there to that describe:

- should throw for invalid type structure provided for named params
- should throw for invalid type structure provided for positional params
- should throw for incorrect number of types provided for positional params

test/bigquery.ts
```diff
@@ -708,26 +735,207 @@ describe('BigQuery', () => {
...
-        return {
-          type: '',
-        };
-      });
+      sandbox
+        .stub(BigQuery, 'getTypeDescriptorFromValue_')
+        .callsFake(value_ => {
+          assert.strictEqual(value_, value);
+          setImmediate(done);
+          return {
```
stephenplusplus: We need to follow this through to verify this object we return is used as we expect.

test/bigquery.ts
```diff
@@ -708,26 +735,207 @@ describe('BigQuery', () => {
...
+        bq.createQueryJob({
+          query: QUERY_STRING,
+          params: POSITIONAL_PARAMS,
+          types: ADDITIONAL_TYPES,
+        });
+      }, /Incorrect number of parameter types provided./);
+      done();
+    });
+
+    it('should throw for empty array without provided types', done => {
```
stephenplusplus: Since this is already tested in the `getTypeDescriptorFromValue_` describe block, we don't need to duplicate here.

test/bigquery.ts
```diff
@@ -708,26 +735,207 @@ describe('BigQuery', () => {
...
+      assert.throws(() => {
+        bq.createQueryJob({
+          query: QUERY_STRING,
+          params: [],
+        });
+      }, /Type must be provided for empty array./);
+      done();
+    });
+
+    it('should throw for null value without provided types', done => {
```
stephenplusplus: Since this is already tested in the `getTypeDescriptorFromValue_` describe block, we don't need to duplicate here.

test/bigquery.ts
steffnay: I'm really curious about this because these tests that check for errors keep timing out if I don't call done(). Is there something else I should be doing to fix that?

test/bigquery.ts
steffnay: Oh, this was just left in by accident.

test/bigquery.ts
stephenplusplus: I think you just need to remove the `done` argument from the test function's argument list:

```diff
- it('should ...', done => {
+ it('should ...', () => {
```

steffnay: Oh, got it.

src/bigquery.ts
```diff
@@ -731,6 +732,63 @@ export class BigQuery extends common.Service {
...
+              providedType[prop]
+            ),
+          };
+        }),
+      };
+    }
+
+    if (!VALID_TYPES.includes(providedType.toUpperCase())) {
+      throw new Error(`Invalid type provided: "${providedType}"`);
+    } else {
```
JustinBeckwith: Nit, but this `else` is not really needed :)

src/bigquery.ts
```diff
@@ -760,10 +820,16 @@ export class BigQuery extends common.Service {
     } else if (value instanceof Big) {
       typeName = 'NUMERIC';
     } else if (is.array(value)) {
-      return {
-        type: 'ARRAY',
-        arrayType: BigQuery.getType_(value[0]),
-      };
+      if (value.length === 0) {
+        throw new Error('Type must be provided for empty array.');
+      } else {
```
JustinBeckwith: same feedback on the else not being needed

src/bigquery.ts
```diff
@@ -760,10 +820,16 @@ export class BigQuery extends common.Service {
...
+        return {
+          type: 'ARRAY',
+          arrayType: BigQuery.getTypeDescriptorFromValue_(value[0]),
+        };
+      }
+    } else if (value === null) {
```
JustinBeckwith: To make branches like this less complex and more readable, I like to pull these to the top, and just do an `if` check that throws.  That way, you need fewer branched `else` clauses.

test/bigquery.ts
steffnay: Okay, I see.

test/bigquery.ts
steffnay: Thanks for catching that.

src/bigquery.ts
```diff
@@ -807,12 +873,18 @@ export class BigQuery extends common.Service {
    * @returns {object} A properly-formed `queryParameter` object.
    */
   // tslint:disable-next-line no-any
-  static valueToQueryParameter_(value: any) {
+  static valueToQueryParameter_(value: any, providedType: any = undefined) {
```
JustinBeckwith: Is `providedType` really `any`?  I'm surprised we can't further constrain this to `{}`.  Side note - for `undefined`, you don't need to set a default value.  You can say `providedType?: {}`

src/bigquery.ts
```diff
@@ -1055,18 +1132,63 @@ export class BigQuery extends common.Service {
...
+            throw new Error(
+              'Provided types must match the value type passed to `params`'
+            );
+          }
+
+          if (query.params.length !== query.types.length) {
+            throw new Error('Incorrect number of parameter types provided.');
+          }
+          // tslint:disable-next-line: no-any
+          query.params.forEach((value: any, i: number) => {
```
JustinBeckwith: Instead of any, guessing you can use `{}`

src/bigquery.ts
```diff
@@ -1055,18 +1132,63 @@ export class BigQuery extends common.Service {
...
+            const queryParameter = BigQuery.valueToQueryParameter_(
+              value,
+              query.types[i]
+            );
+            query.queryParameters.push(queryParameter);
+          });
+        } else {
+          // tslint:disable-next-line: no-any
+          query.params.forEach((value: any) => {
```
JustinBeckwith: For most uses of `any` here, `{}` would do

system-test/bigquery.ts
```diff
@@ -1035,6 +1035,21 @@ describe('BigQuery', () => {
             );
           });
 
+          it('should work with empty arrays', done => {
```
JustinBeckwith: This could be far shorter if done with async style:

```js
it('should work with empty arrays', async () => {
  const [rows] = await bigquery.query({
    query: 'SELECT * FROM UNNEST (?)',
    params: [[]],
    types: [['INT64']],
  });
  assert.strictEqual(rows.length, 0);
});
```

For new code, I would prefer we avoid callbacks, and lean on async/await style code

system-test/bigquery.ts
```diff
@@ -1270,6 +1285,23 @@ describe('BigQuery', () => {
...
```
JustinBeckwith: same feedback as above

test/bigquery.ts
```diff
@@ -708,26 +735,207 @@ describe('BigQuery', () => {
...
+        BigQuery.getTypeDescriptorFromValue_(undefined);
+      }, expectedError);
+    });
+
+    it('should throw with an empty array', () => {
+      const expectedError = new RegExp(
+        'Type must be provided for empty array.'
+      );
+
+      assert.throws(() => {
```
JustinBeckwith: Just a style thing, but you don't need to define the `RegExp` out of band like this.  You can say:

```js
assert.throws(
  ()=> BigQuery.getTypeDescriptorFromValue_([]),
  /Type must be provided for empty array/
);
```

test/bigquery.ts
```diff
@@ -708,26 +735,207 @@ describe('BigQuery', () => {
...
+            name: 'prop',
+            type: {
+              type: 'INT64',
+            },
+          },
+        ],
+      });
+    });
+
+    it('should throw for invalid provided types with named params', done => {
```
JustinBeckwith: Let's use async instead of callbacks (no `done` for new code)

src/bigquery.ts
steffnay: providedType can be a string, object, or array, so I don't understand how to make providedType?: {} work this way.

system-test/bigquery.ts
steffnay: Okay, I'll start using async/await instead.

src/bigquery.ts
bcoe: you could do something along the lines of:

```
providedType: string|{[key: string]: string}|string[]
```

:point_up: `|` is used to allow for multiple types.

steffnay: @bcoe The array type can also be populated with string, array, or object. How do I specify that without using any?

bcoe: @steffnay you'd potentially probably want to start breaking it out into interfaces, but you'd do something along the lines of:

```
{[key: string]: string|string[]}
```

:point_up: but as it gets complex like this, you'd probably want to start describing an interface. The advantage of putting in this work is that folks using the API have more safety, and it's easier to figure out what to pass to the API.

PR #647, opened 2020-03-17
refactor!: don't return Stream from createLoadJob

### LINKED ISSUE #640: table.createLoadJob has mutually exclusive polymorphic return
Affects `@google-cloud/bigquery@4.7.0`

The `table.createLoadJob` is a callback-first method that has been promisified via `promisifyAll`, i.e. it doesn't meet the requirements of a non-promise method (`Stream`/`constructor`/etc). However, when the `source` argument is a `string`, the current implementation attempts a synchronous return type of `Stream.Writable` (`EventEmitter`). See https://github.com/googleapis/nodejs-bigquery/blob/v4.7.0/src/table.ts#L1226-L1248

This polymorphic return is problematic:

1. With callback and without await, Writable is returned
2. With callback and with await, Writable is returned
3. Without callback and without await, Promise (of Job? can't actually test this) is returned
4. Without callback and with await, Promise of Job is returned

The corresponding overloads are available:
```ts
createLoadJob(source: string): Writable; // Never reached
createLoadJob(source: string, callback: JobCallback): Writable; // 1, 2

// 4 is not documented as an overload, but is possible in the concrete signature:
createLoadJob(source: string): Promise<JobResponse> {...}
```

One of the tests expects a `Writable` without a supplied callback, but, upon promisifyAll refactor in the unit tests to actually be applied, it fails due to the method assuming to be used like a promise, which corresponds to scenario 3 in previous: https://github.com/googleapis/nodejs-bigquery/blob/v4.7.0/test/table.ts#L58-L64

In the current unit tests, this issue is not presented due to the promisifyAll stub, which removes the promisification functionality. See https://github.com/googleapis/nodejs-bigquery/blob/v4.7.0/test/table.ts#L58-L64

Example test cases (all succeed):
```ts
// fix the promisifyAll fake
const fakePfy = extend({}, pfy, {
  promisifyAll: (c: Function) => {
    if (c.name === 'Table') {
      promisified = true;
    }
    pfy.promisifyAll(c); // ADDED THIS
  },
});

// test cases
it('+cb +async', async () => {
  sandbox.stub(table, 'createWriteStream_').returns(new stream.Writable());
  const result = await table.createLoadJob(FILEPATH, () => {}); // noop
  assert(result instanceof stream.Writable);
  result.emit('job', {metadata: 'foo'});
});

it('+cb -async', (done) => {
  sandbox.stub(table, 'createWriteStream_').returns(new stream.Writable());
  const result = table.createLoadJob(FILEPATH, done);
  assert(result instanceof stream.Writable);
  result.emit('job', {metadata: 'bar'});
});

it('-cb +async', async () => {
  const w = new stream.Writable();
  sandbox.stub(table, 'createWriteStream_').returns(w);
  setImmediate(() => w.emit('job', {metadata: 'qux'}));
  const result = await table.createLoadJob(FILEPATH);
  assert.deepStrictEqual(result, [{metadata: 'qux'}, 'qux']);
});

it('-cb -async', () => {
  // cannot test this, table.createLoadJob always returns a Promise
});
```

Typically, combining promises, callbacks, _and_ streams in a single function signature is mutually exclusive.

If a `Writable` result is really desired that coincides with the `createLoadJob(source: string): Writable;` signature, I recommend implementing a `createLoadJobStream(source: string): Writable;` method which will _not_ be promisified.

zamnuts:
Breaking change to fix #640:

- `createLoadJob` now only returns a promise (or handles the callback), and provides the completed job response. in some signatures this used to return a Writable stream
- added `createLoadJobStream` to explicitly return the Writable without any async operations, unless the writable is specifically desired by the consumer, `createLoadJob` is recommended.

fix!(table): createLoadJobStream sync returns a stream, createLoadJob always returns a job #640

- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Appropriate docs were updated (if necessary)

Fixes #640 🦕

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -20,6 +20,7 @@ import {promisifyAll} from '@google-cloud/promisify';
 import arrify = require('arrify');
 import Big from 'big.js';
 import * as extend from 'extend';
+import pEvent from 'p-event';
 
 const format = require('string-format-obj');
 import * as fs from 'fs';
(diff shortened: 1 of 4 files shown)


src/table.ts
```diff
@@ -1203,17 +1200,31 @@ class Table extends common.ServiceObject {
...
+  ): void | Promise<JobResponse> {
     const metadata =
       typeof metadataOrCallback === 'object' ? metadataOrCallback : {};
     const callback =
-      typeof metadataOrCallback === 'function'
-        ? metadataOrCallback
-        : cb || common.util.noop;
+      typeof metadataOrCallback === 'function' ? metadataOrCallback : cb;
+
+    this._createLoadJob(source, metadata).then(
```
stephenplusplus: Can we bring the functionality from `_createLoadJob` into this function?

zamnuts: that's how it was before, i did this to minimize callback usage in the main body of the logic.

src/table.ts
```diff
@@ -1238,13 +1249,11 @@ class Table extends common.ServiceObject {
       }
 
       // Read the file into a new write stream.
-      return fs
+      const jobWritable = fs
         .createReadStream(source)
-        .pipe(this.createWriteStream_(metadata))
-        .on('error', callback)
```
stephenplusplus: What will happen now if this stream errors?

zamnuts: promise rejection (throw). `pEvent` defaults to listening for `error` event, and rejects the promise.

bcoe:
lets land in the next major, this looks great to me though.

PR #687, opened 2020-04-27
docs(samples): Adds Samples

### LINKED ISSUE #328: table.insert should (recursively) convert from native types to correct JSON

It's frustrating to determine the correct JSON format for each data type when inserting a row with table.insert. For example, if I have a Date object and want to insert it as a TIMESTAMP value in BigQuery, how would I do that? Also, if encoding large integers, I'd have to send as a string (though it's true that JavaScript requires extra modules to support int64)

The Python client's insert_rows method converts from native Python type to the correct JSON. It does this recursively so that array and struct columns are also supported.

https://github.com/googleapis/google-cloud-python/blob/64d1dc857f876cc69dcd21e88efa255cf5d3debd/bigquery/google/cloud/bigquery/client.py#L1497

The relevant converters are at

https://github.com/googleapis/google-cloud-python/blob/64d1dc857f876cc69dcd21e88efa255cf5d3debd/bigquery/google/cloud/bigquery/_helpers.py#L311-L436

Alternative: for each column type, document the precise encoding expected (and a helper function for how to convert it).

 **Additional context**

This is in response to @jasonpolites's helpful friction log.

steffnay:
Adds Samples:
- Authorized view tutorial
- Integer range partitioning
- Querying external tables
- Inserting supported data types

Adds region tag to relaxing column sample
Fixes #328 🦕


samples/createTableRangePartitioned.js
```diff
@@ -0,0 +1,67 @@
...
+    // const tableId = "my_table";
+
+    const schema = [
+      {name: 'fullName', type: 'STRING'},
+      {name: 'city', type: 'STRING'},
+      {name: 'zipcode', type: 'INTEGER'},
+    ];
+
+    // To use integer range partitioning, select a top-level REQUIRED or
+    // NULLABLE column with INTEGER / INT64 data type.
```
shollyman, line 40: nit:  should we indicate what happens to values outside the range?  Or is there a link in the docs we can send people to get more information here?

samples/insertingDataTypes.js
```diff
@@ -0,0 +1,140 @@
...
+    // independent of a specific date and timezone.
+    const bqTime = bigquery.time('14:00:00');
+    // A TIMESTAMP object represents an absolute point in time,
+    // independent of any time zone or convention such as Daylight
+    // Savings Time with microsecond precision.
+    const bqTimestamp = bigquery.timestamp('2020-04-27T18:07:25.356Z');
+    const bqGeography = bigquery.geography('POINT(1 2)');
+    const schoolBuffer = Buffer.from('Test University');
+
+    // Rows to be inserted into table
```
shollyman, line 112: nit:  maybe clarify you can send one or more rows, given its calls rows but only contains a single row here?  Or send a second row that only have a few explicit fields to show batching more clearly?  I do like the use of complex types here though.

samples/queryExternalGCSPerm.js
```diff
@@ -0,0 +1,79 @@
...
+      csvOptions: {skipLeadingRows: 1},
+    };
+
+    // For all options, see https://cloud.google.com/bigquery/docs/reference/v2/tables#resource
+    const options = {
+      schema: schema,
+      externalDataConfiguration: dataConfig,
+    };
+
+    // Create a permanent table linked to the GCS file
```
shollyman, line 53: nit:  maybe call this a federated or an external table rather than a permanent one?

samples/test/authViewTutorial.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+const {assert} = require('chai');
+const {describe, it, before, after} = require('mocha');
+const cp = require('child_process');
+const uuid = require('uuid');
+
+const {BigQuery} = require('@google-cloud/bigquery');
+
+const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
+const generateUuid = () => `gcloud-tests-${uuid.v4()}`.replace(/-/gi, '_');
```
shollyman, line 25: maybe give these a more descriptive uuid prefix so you can find what's creating them if things go badly and they don't get cleaned up?   gcloud-tests seems unnecessarily vague (and would have me seeking out something in the gcloud sdk if I saw them).

PR #699, opened 2020-05-13
refactor: reduce usage of extend

JustinBeckwith:


test/bigquery.ts
```diff
@@ -401,7 +400,7 @@ describe('BigQuery', () => {
         },
       ];
 
-      const schemaObject = extend(true, SCHEMA_OBJECT, {});
+      const schemaObject = Object.assign(SCHEMA_OBJECT, {});
```
stephenplusplus, line 403: Without using the deep clone, we are just modifying "SCHEMA_OBJECT". It is harmless because this happens to be the last test in the describe block, but if it was shifted or something new was added, it'd be a weird thing to track down. If we're not outright ditching `extend`, could we keep it in cases where we use a deep clone? Or maybe we need a "teeny-extend"? These overhauls are scary because the changes are very subtle, and we might not be the ones to notice the effects.

JustinBeckwith: Just to be clear, this PR is super not necessary 😆 I like to use native features where I can, and if there are points we're using them incorrectly, the answer would be for me to back it out.  

Now - the reason I tried changing this one.  My understanding of the `deep` boolean passed into extend is that it will perform a clone of objects down the chain as opposed to cloning references (I am admittedly fuzzy here).  I have no clue what this original line of code was trying to do, because it looks like it's merging an empty object into an existing one.

What was the desired intent here / what am I missing?

stephenplusplus: In this case, "SCHEMA_OBJECT" is meant to be used throughout multiple "its", but this test needs to change the contents of the `fields` array. The shallow clone in this case will only give us a new shell object that we can modify, but when it hits an array or object, it just references it.

>  const schemaObject = extend(true, SCHEMA_OBJECT, {});

Yeesh, that was wrong apparently. We wanted "extend(true, {}, SCHEMA_OBJECT)". That would allow us to not modify the "fields" array from the "global" `SCHEMA_OBJECT` variable.

JustinBeckwith: Thank you for explaining this :) So think you're saying this:

```js
Object.assign({}, { a: { b: [0, 1, 2] } })`
```

Is not the same as:

```js
extend(true, {}, { a: { b: [0, 1, 2] } });
```

Largely because in the `Object.assign` case, it would copy a reference to the array into the newly formed object, and in the `extend` case it will fully clone the array.  Am I getting it?

stephenplusplus: Yep, exactly.

```js
const originalObj = { things: [1, 2, 3] }
const assignClone = Object.assign({}, originalObj)
const extendClone = extend(true, {}, originalObj)

assignClone.uniqueProperty = true
console.log(assignClone.uniqueProperty) // true
console.log(originalObj.uniqueProperty) // undefined

assignClone.things.push(4)
console.log(originalObj.things === assignClone.things)
console.log(originalObj.things) // [1, 2, 3, 4]

// vs...

extendClone.things.push(4)
console.log(originalObj.things === extendClone.things) // false
console.log(originalObj.things) // [1, 2, 3]
console.log(extendClone.things) // [1, 2, 3, 4]
```

PR #712, opened 2020-06-07
fix: selectedFields on getRows not working correctly 

### LINKED ISSUE #702: selectedFields on getRows not working correctly
#### Environment details
  - `@google-cloud/bigquery` version:
    "@google-cloud/bigquery": "^4.7.0",

#### Steps to reproduce

  1. create a bigquery client, call getRows() passing in selectedFields into the options
  2. depending on what selectedFields are used it can return invalid data that doesnt match the selectedFields passed in. sometimes it will return a different field, or it will return a value from a column that isnt in the one passed into selectedFields. 

example  - here is my table 
```
==========================================
|season | round | days_from_epoch |.... other columns
==========================================
```

if I pass in **season** as the selectedFields, i get the right result 
```
[
  {
    "season": 1989
  },
  {
    "season": 1997
  },
  {
    "season": 1997
  },
  {
    "season": 1994
  },
  {
    "season": 1985
}
]
```
but if I pass in **round** for selectedFields, i get results where the _key_ is season but the value is from the _round_  column, that doesnt make sense
```
[
  {
    "season": 64
  },
  {
    "season": 32
  },
  {
    "season": 64
  },
  {
    "season": 64
  },
  {
    "season": 64
  },
```

if i pass in **round,days_from_epoch** for selectedFields, i get season and round back, which doesn't make sense - in this case the data is right, but i never even passed in 'season' as a selectedField
```
[
  {
    "season": 64,
    "round": 7015
  },
  {
    "season": 32,
    "round": 9936
  },
  {
    "season": 64,
    "round": 9934
  },
  {
    "season": 64,
    "round": 8842
  },
]
```

i tried this with the bq CLI tool, and i got the right results for selectedFields for all the above test cases, so I think it is something involving this library. 

```
bq head --max_rows 3 --selected_fields "season" 
+--------+
| season |
+--------+
|   1989 |
|   1997 |
|   1997 |
+--------+

bq head --max_rows 3 --selected_fields "round" 
+-------+
| round |
+-------+
|    64 |
|    32 |
|    64 |
+-------+
bq head --max_rows 3 --selected_fields "season,round" 
+--------+-------+
| season | round |
+--------+-------+
|   1989 |    64 |
|   1997 |    32 |
|   1997 |    64 |
+--------+-------+
bq head --max_rows 3 --selected_fields "round,days_from_epoch"
+-------+-----------------+
| round | days_from_epoch |
+-------+-----------------+
|    64 |            7015 |
|    32 |            9936 |
|    64 |            9934 |
+-------+-----------------+

```

here's my code around the BigQuery client, im pretty sure it's not a bug in here, i have a whole bunch of other integration code that appears to work correctly. 
```
			const bigquery = getBigQueryClient()
			// For all options, see https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query
			const selectedFields = req.query.selectedFields
			console.log("Selected fields", selectedFields)
			const options = {
				// Location must match that of the dataset(s) referenced in the query.
				location: 'US',
				maxResults:100,
				selectedFields
			};
			const dataset = req.query.dataset
			const table = req.query.table

			// Run the query as a job
			bigquery.dataset(dataset).table(table).getRows(options).then((rows) => {
				console.log("got rows", rows)
				res.json(rows[0])
			}).catch( e=> res.json(getError(e)))
```

laljikanjareeya:
Fixes #702 

Data from Back-end is retrieving correctly, but mapping of the data with the schema fields has an issue.
On below line schema fields and data mapping is based on index so, it will map always first property even if user doesn't selected that property.

https://github.com/googleapis/nodejs-bigquery/blob/93945dd30985143ed45472e8744f30b2f2a17793/src/bigquery.ts#L387

Solution:
Back-end data is retrieving same order as schema fields order, solution for this is to filter schema based on selected fields when `selectedFields` is provided by user.


steffnay:
Looks good to me. Pending second review from @stephenplusplus

src/bigquery.ts
```diff
@@ -376,12 +376,34 @@ export class BigQuery extends common.Service {
    *
    * @param {object} schema
    * @param {array} rows
+   * @param {string} selectedFields List of fields to return (comma-separated).
```
stephenplusplus, line 379: Can we accept this as an array of strings instead?

laljikanjareeya: Done!!

laljikanjareeya:
@googlebot I fixed it.

PR #717, opened 2020-06-09
tests(system): updates tests to identify & delete stale dataset resources

steffnay:

Fixes #<715 🦕

### DIFF
--- a/system-test/bigquery.ts
+++ b/system-test/bigquery.ts
@@ -1673,7 +1673,11 @@ describe('BigQuery', () => {
   function isResourceStale(creationTime: number | string) {
     const oneDayMs = 86400000;
     const now = new Date();
-    const created = new Date(creationTime);
+    let created = new Date(creationTime);
+    // If creationTime was returned as unix timestamp, convert.
+    if (created.toString() === 'Invalid Date') {
+      created = new Date(parseInt(creationTime as string));
+    }
     return now.getTime() - created.getTime() >= oneDayMs;
   }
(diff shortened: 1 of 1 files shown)


stephenplusplus:
I was able to only make this change from master and it seemed to be working: 

```diff
diff --git a/system-test/bigquery.ts b/system-test/bigquery.ts
index fe7007e..b0e2d36 100644
--- a/system-test/bigquery.ts
+++ b/system-test/bigquery.ts
@@ -1699,9 +1699,9 @@ describe('BigQuery', () => {
     });
 
     const deleteDatasetPromises = datasets
-      .filter(dataset => {
-        const creationTime = dataset.metadata.creationTime;
-        return creationTime && isResourceStale(creationTime);
+      .filter(async dataset => {
+        const [metadata] = await dataset.getMetadata();
+        return isResourceStale(parseInt(metadata.creationTime));
       })
       .map(dataset => {
         return dataset.delete({force: true});
```

Good catch on the `dataset.metadata` object not holding the `creationTime` property.

I was looking at `creationTime` as documented [here](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#Dataset), and it says

> **string (int64 format)**
>
> Output only. The time when this dataset was created, in milliseconds since the epoch.

Is it true it could be undefined or something else?

steffnay:
@stephenplusplus Sweet, that's so much easier. creationTime should never be undefined. I think this fixes the issue.

PR #748, opened 2020-06-11
fix: load job to a different project ID

### LINKED ISSUE #719: Unable to run load job to a different project ID when using local file
It appears the approach mentioned in #26 does not work with the current version of the library (4.7.0). That is, the code below does not appear to properly change the `projectId` of the destination table.
```js
table.load('test.csv', {
  destinationTable: {
    projectId: 'test-override'
  }
})
```

Looking at the implementation https://github.com/googleapis/nodejs-bigquery/blob/92599f98bf50b2c0e4daf1082e05454478d3c6bb/src/table.ts#L1376-L1382
it looks like the `extend` is overwriting any user provided metadata around the `destinationTable`. Should that be reversed?

laljikanjareeya:
Fixes #719

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -1373,16 +1373,17 @@ class Table extends common.ServiceObject {
       metadata.schema = Table.createSchemaFromString_(metadata.schema);
     }
 
-    metadata = {
-      ...{
+    metadata = extend(
+      true,
+      {
         destinationTable: {
           projectId: this.bigQuery.projectId,
           datasetId: this.dataset.id,
           tableId: this.id,
         },
       },
-      ...metadata,
-    };
+      metadata
+    );
 
     let jobId = metadata.jobId || uuid.v4();
 
--- a/src/table.ts
+++ b/src/table.ts
@@ -1373,13 +1373,16 @@ class Table extends common.ServiceObject {
       metadata.schema = Table.createSchemaFromString_(metadata.schema);
     }
 
-    extend(true, metadata, {
-      destinationTable: {
-        projectId: this.bigQuery.projectId,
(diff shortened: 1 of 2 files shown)


src/table.ts
```diff
@@ -1373,13 +1373,16 @@ class Table extends common.ServiceObject {
       metadata.schema = Table.createSchemaFromString_(metadata.schema);
     }
 
-    extend(true, metadata, {
-      destinationTable: {
-        projectId: this.bigQuery.projectId,
-        datasetId: this.dataset.id,
-        tableId: this.id,
+    metadata = {
```
stephenplusplus, line 1376: I don't believe this style will do a deep clone of the object:

```js
const defaultConfig = { deep: { obj: { value: false }}}
const userConfig = { deep: { obj: { value: true }}}

// Override defaults with user config
const metadata = {...defaultConfig, ...userConfig}

// Confirm the overwrite worked...
console.log(metadata.deep.obj.value) // true (good!)

// Set a new property on the clone object...
metadata.deep.obj.newValue = false

// Check if it modified the user's config object
console.log(userConfig.deep.obj.newValue) // false (d'oh! The user's object should not be modified)
```

A fix for this would be sticking to `extend`:

```diff
// Override defaults with user config
-const metadata = {...defaultConfig, ...userConfig}
+const metadata = extend(true, defaultConfig, userConfig)
```

laljikanjareeya: @stephenplusplus Thanks!! I was using same solution that you provided but was not reassign to `metadata` so it wasn't extend the existing object. but now it's worked.

stephenplusplus: Thank you!

PR #749, opened 2020-06-11
doc: docs for defaultDataset

### LINKED ISSUE #633: defaultDataset option not documented
Re: https://googleapis.dev/nodejs/bigquery/latest/BigQuery.html#createQueryJob:

It does not mention the ability to define a default dataset for the query.

However, code can do things like this:

```
  const query = {query: queryString};
  if (connection.datasetName) {
    query.defaultDataset = { datasetId: connection.datasetName };
  }
  // ...
  return bigquery
    .createQueryJob(query)
    .then(...);
```

If I should report this via the Google Issue Tracker, let me know what component to assign it to. I couldn't figure it out. In any case, I thought the report would do more practical good here.

laljikanjareeya:
Fixes #633


src/bigquery.ts
```diff
@@ -90,7 +90,6 @@ export type Query = JobRequest<bigquery.IJobConfigurationQuery> & {
   params?: any[] | {[param: string]: any};
   dryRun?: boolean;
   types?: string[] | string[][] | {[type: string]: string[]};
-  defaultDataset?: Dataset;
```
JustinBeckwith, line 93: Why did the property get deleted from the interface?

laljikanjareeya: it's already there in `IJobConfigurationQuery`.
types are different one is `Dataset` and another one is `IDatasetReference`.

steffnay (approved):

PR #752, opened 2020-06-11
tests(system): adds check for creationTime

steffnay:

 🦕

### DIFF
--- a/system-test/bigquery.ts
+++ b/system-test/bigquery.ts
@@ -1701,7 +1701,8 @@ describe('BigQuery', () => {
     const deleteDatasetPromises = datasets
       .filter(async dataset => {
         const [metadata] = await dataset.getMetadata();
-        return isResourceStale(parseInt(metadata.creationTime));
+        const creationTime = parseInt(metadata.creationTime);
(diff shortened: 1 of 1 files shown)


system-test/bigquery.ts
```diff
@@ -1701,7 +1701,8 @@ describe('BigQuery', () => {
     const deleteDatasetPromises = datasets
       .filter(async dataset => {
         const [metadata] = await dataset.getMetadata();
-        return isResourceStale(parseInt(metadata.creationTime));
+        const creationTime = parseInt(metadata.creationTime);
```
bcoe, line 1704: In what instances is this not set, out of curiosity?

JustinBeckwith, line 1704: You almost always want to use `Number(metadata.creationTime)` over `parseInt`.  ParseInt has some weird behaviors around radix detection and string parsing.  This thread has a great explanation:
https://stackoverflow.com/questions/4090518/what-is-the-difference-between-parseint-and-number

steffnay: This is really helpful, thank you!

PR #754, opened 2020-06-12
test: refactors system test deleteDatasets()

steffnay:
This refactor adds logic to check the existence of a stale testing dataset before attempting to delete it. This should fix tests that are failing due to parallel CI test runs attempting to delete the same resource.

 🦕

### DIFF
--- a/system-test/bigquery.ts
+++ b/system-test/bigquery.ts
@@ -1698,23 +1698,18 @@ describe('BigQuery', () => {
       filter: `labels.${GCLOUD_TESTS_PREFIX}`,
     });
 
-    const datasetTimePromises = datasets.map(async dataset => {
+    for (const dataset of datasets) {
       const [metadata] = await dataset.getMetadata();
-      return [dataset, metadata.creationTime];
-    });
+      const creationTime = Number(metadata.creationTime);
 
-    const datasetTimes = await Promise.all(datasetTimePromises);
-
-    const deleteDatasetPromises = datasetTimes
-      .filter(datasetTime => {
-        const creationTime = Number(datasetTime[1]);
-        return isResourceStale(creationTime);
-      })
-      .map(async datasetTime => {
-        const [exists] = await datasetTime[0].exists();
-        return exists ? datasetTime[0].delete({force: true}) : undefined;
-      });
-
-    await Promise.all(deleteDatasetPromises);
+      if (isResourceStale(creationTime)) {
+        try {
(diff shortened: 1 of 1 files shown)


system-test/bigquery.ts
```diff
@@ -101,11 +101,13 @@ describe('BigQuery', () => {
   ];
 
   before(async () => {
-    // Remove buckets created for the tests.
-    await deleteBuckets();
+    await Promise.all([
```
bcoe, line 104: I'm surprised this would cause flakiness, the difference between these two approaches would be that when you go:

```js
await deleteBuckets();
await deleteDatasets();
```

delete buckets always happens before delete datasets.

Whereas, with:

```js
Promise.all([
  deleteBuckets(),
  deleteDatasets()
])
```

It's going to try to delete in parallel, scratching my head 🤔

system-test/bigquery.ts
```diff
@@ -1698,14 +1698,21 @@ describe('BigQuery', () => {
       filter: `labels.${GCLOUD_TESTS_PREFIX}`,
     });
 
-    const deleteDatasetPromises = datasets
-      .filter(async dataset => {
-        const [metadata] = await dataset.getMetadata();
-        const creationTime = Number(metadata.creationTime);
-        return creationTime && isResourceStale(creationTime);
+    const datasetTimePromises = datasets.map(async dataset => {
+      const [metadata] = await dataset.getMetadata();
```
stephenplusplus, line 1702: I think we may have slipped a bug in here-- I don't think "await" will actually work in an Array iterator like `map`/`forEach`/`filter`. The array iterator will return immediately and the promises will resolve later, but without a guarantee before we move on to the next code that depends on the values.

I tried rewriting this function with a `for of`, and I think this will do what we want:

```js
async function deleteDatasets() {
  const [datasets] = await bigquery.getDatasets({
    filter: `labels.${GCLOUD_TESTS_PREFIX}`,
  });

  for (const dataset of datasets) {
    const [metadata] = await dataset.getMetadata();
    const creationTime = Number(metadata.creationTime);

    if (isResourceStale(creationTime)) {
      try {
        await dataset.delete({force: true});
      } catch (e) {
        console.log(`dataset(${dataset.id}).delete() failed`);
        console.log(e);
      }
    }
  }
}
```

Note that I changed the call for "exists()" to just console.log the error if we get one.

steffnay: Okay, thanks!

PR #797, opened 2020-06-18
docs(samples): adds sample for providing query parameter types

### LINKED ISSUE #703: Provide Documentation on Parameter Types

BigQuery supports parameters.  However, if you pass in a null parameter, it will error out complaining that it's unable to figure out the type.  

This was fixed, the nodejs client now supports passing in a second type array/object to provide types.  However, this is poorly documented. For example, even after knowing all this, we ran into an issue of not knowing what type to pass in for an Array of Strings.

The BigQuery nodejs documentation should include information on how to pass in type, and examples.  This should cover all types, including arrays and structs.

It may also be useful for the error message to link said docs.

steffnay:

Fixes #703 🦕

### DIFF
--- a/samples/queryParamsNamedTypes.js
+++ b/samples/queryParamsNamedTypes.js
@@ -30,7 +30,7 @@ function main(
     /**
      * TODO(developer): Uncomment the following lines before running the sample.
      */
-    // projectId = process.env.PROJECT_ID;
+    // projectId = 'my_project';
     // const datasetId = 'my_dataset';
     // const tableId = 'my_table';
 
--- a/samples/queryParamsNamedTypes.js
+++ b/samples/queryParamsNamedTypes.js
@@ -1,0 +1,81 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


samples/queryParamsNamedTypes.js
```diff
@@ -0,0 +1,81 @@
...
+  const {BigQuery} = require('@google-cloud/bigquery');
+  const bigquery = new BigQuery();
+
+  async function queryParamsNamedTypes() {
+    // Run a query using named query parameters and provided parameter types.
+
+    /**
+     * TODO(developer): Uncomment the following lines before running the sample.
+     */
+    // projectId = process.env.PROJECT_ID;
```
JustinBeckwith, line 33: In the stye guide, we use a hard coded string for projectId:
https://g3doc.corp.google.com/company/teams/cloud-devrel/dpe/samples/nodejs.md?cl=head

samples/queryParamsPositionalTypes.js
```diff
@@ -0,0 +1,81 @@
...
+  const {BigQuery} = require('@google-cloud/bigquery');
+  const bigquery = new BigQuery();
+
+  async function queryParamsPositionalTypes() {
+    // Run a query using positional query parameters and provided parameter types.
+
+    /**
+     * TODO(developer): Uncomment the following lines before running the sample.
+     */
+    // projectId = process.env.PROJECT_ID;
```
JustinBeckwith, line 33: Same issue with projectId - I don't think that's a env var we ask folks to set

samples/queryParamsNamedTypes.js
steffnay: Okay, got it! But now I need to figure out where I saw this as an example and update it...

PR #801, opened 2020-06-18
fix: blunderbuss config

sofisl:


.github/blunderbuss.yml
```diff
@@ -1,6 +1,8 @@
 assign_issues:
   - sofisl
   - bcoe
```
bcoe, line 3: @sofisl depending on @steffnay's direction, it might make sense to remove you and I, and add @stephenplusplus?

@steffnay who would you like to help triage on this repo?

steffnay: Yeah, @stephenplusplus would be really helpful.

PR #813, opened 2020-07-14
feat: detect param type if not in provided types

### LINKED ISSUE #802: Allow option to provide only some named parameter types
Currently, if a user wants to provide a type for a named query parameter, they are required to provide types for all parameters. An improvement would be for the library to default to using a provided type, but fall back on inferring the type if one hasn't been provided.

steffnay:
Adds ability to detect named parameter type if types were provided for some parameters, but not all.

Fixes #802 🦕


bcoe (approved):

stephenplusplus (approved):
LGTM. A system test could be useful here, too.

shollyman (approved):

pdfabbro:
Was this actually applied?  I'm having an issue right now where it's saying I'm not providing the type for an empty array even though I am and I'm starting to wonder if it's because I'm not providing the type for all the other params too

PR #830, opened 2020-07-29
docs(samples): adds routine and user agent samples

steffnay:
- Adds samples for get/list/update/delete routines 

- Adds sample for setting provided user agent

- Updates samples dataset tests to delete stale datasets

- Refactors samples tests resource naming

🦕

### DIFF
--- a/samples/createRoutine.js
+++ b/samples/createRoutine.js
@@ -58,7 +58,7 @@ function main(
 
     console.log(`Routine ${routineId} created.`);
   }
-  // [END bigquery_create_routine]
   createRoutine();
+  // [END bigquery_create_routine]
 }
 main(...process.argv.slice(2));
--- a/samples/createRoutine.js
+++ b/samples/createRoutine.js
@@ -1,0 +1,64 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+function main(
+  datasetId = 'my_dataset', // Existing dataset
+  routineId = 'my_routine' // Routine to be created
+) {
+  // [START bigquery_create_routine]
+  // Import the Google Cloud client library and create a client
+  const {BigQuery} = require('@google-cloud/bigquery');
+  const bigquery = new BigQuery();
+
+  async function createRoutine() {
+    // Creates a new routine named "my_routine" in "my_dataset".
+
+    /**
+     * TODO(developer): Uncomment the following lines before running the sample.
+     */
+    // const datasetId = 'my_dataset';
+    // const routineId = 'my_routine';
+
+    const dataset = bigquery.dataset(datasetId);
+
+    // Create routine reference
+    let routine = dataset.routine(routineId);
+
+    const config = {
+      arguments: [
+        {
+          name: 'x',
+          dataType: {
+            typeKind: 'INT64',
+          },
+        },
+      ],
+      definitionBody: 'x * 3',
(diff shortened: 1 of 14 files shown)


samples/createRoutine.js
```diff
@@ -0,0 +1,64 @@
...
+      },
+    };
+
+    // Make API call
+    [routine] = await routine.create(config);
+
+    console.log(`Routine ${routineId} created.`);
+  }
+  // [END bigquery_create_routine]
+  createRoutine();
```
JustinBeckwith, line 62: This needs to be inside the snippet region tag

samples/createRoutineDDL.js
```diff
@@ -0,0 +1,52 @@
...
+    const [job] = await bigquery.createQueryJob(options);
+    console.log(`Job ${job.id} started.`);
+
+    // Wait for the query to finish
+    await job.getQueryResults();
+
+    console.log(`Routine ${routineId} created.`);
+  }
+  // [END bigquery_create_routine_ddl]
+  createRoutineDDL();
```
JustinBeckwith, line 50: Here and below, need these inside the region tags

samples/getRoutine.js
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * TODO(developer): Uncomment the following lines before running the sample.
+     */
+    // const datasetId = 'my_dataset';
+    // const routineId = 'my_routine';
+
+    const dataset = bigquery.dataset(datasetId);
+
+    // Create routine reference
+    let routine = dataset.routine(routineId);
```
JustinBeckwith, line 38: Could we shorten this to:
```js
const [routine] = await dataset.routine(routineId).get();
```

samples/setUserAgent.js
```diff
@@ -0,0 +1,31 @@
...
+'use strict';
+
+function main() {
+  // [START bigquery_set_user_agent]
+  // Import the Google Cloud client library
+  const {BigQuery} = require('@google-cloud/bigquery');
+  // Create a client and set the user agent
+  const bigquery = new BigQuery({userAgent: 'my-user-agent'});
+
+  async function setUserAgent() {
```
JustinBeckwith, line 24: So this sample has no `await` in it - so no need to have the wrapping async function!  You can drop the `async function setUserAgent` wrapping.

samples/test/routines.test.js
```diff
@@ -0,0 +1,132 @@
...
+const newRoutineId = generateUuid();
+
+const bigquery = new BigQuery();
+
+describe('Routines', () => {
+  after(async () => {
+    await bigquery
+      .dataset(datasetId)
+      .delete({force: true})
+      .catch(console.warn);
```
JustinBeckwith, line 38: Do we have any concerns with leaking datasets here?  I'm ok with the `console.warn`, it's just unlikely it will ever be seen unless running locally.  What's the chance of leakage?

steffnay: @JustinBeckwith I think there's a really low chance of this failing. But we would leak datasets if the tests were stopped mid-run. An option to have a failed delete retried at some point could be to add labels to these testing datasets, then add logic to the before hook to find and delete any stale datasets with the label. This is the pattern used in the system tests and I think it makes sense to just add it to all the samples tests too.

steffnay: I added `deleteDatasets()` to the datasets test before hook, so leaked datasets should be cleaned up on subsequent runs.

samples/test/datasets.test.js
```diff
@@ -105,4 +111,34 @@ describe('Datasets', () => {
     const [exists] = await bigquery.dataset(datasetId).exists();
     assert.strictEqual(exists, false);
   });
+
+  // Only delete a resource if it is older than 24 hours. That will prevent
+  // collisions with parallel CI test runs.
+  function isResourceStale(creationTime) {
+    const oneDayMs = 86400000;
```
JustinBeckwith, line 118: Non-blocking feedback, I like to spell this out as `24 * 60 * 60 * 1000` to make it easier to reason about

samples/test/datasets.test.js
```diff
@@ -105,4 +111,34 @@ describe('Datasets', () => {
...
+    const now = new Date();
```
JustinBeckwith, line 119: Another approach you could take here is to use `Date.now()`.  If `creationTime` is  a "milliseconds since bleh" kind of value:
```js
return Date.now() - creationTime >= oneDayMs
```

JustinBeckwith (approved):

PR #832, opened 2020-07-30
feat: model extract

steffnay:
Adds methods for extracting a model resource to Google Cloud Storage by creating an extract job.

 🦕

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -230,30 +230,27 @@ class Model extends common.ServiceObject {
   }
 
   createExtractJob(
-    destination: File | File[] | string | string[],
+    destination: string | File,
     options?: CreateExtractJobOptions
   ): Promise<JobResponse>;
   createExtractJob(
-    destination: File | File[] | string | string[],
+    destination: string | File,
     options: CreateExtractJobOptions,
     callback: JobCallback
   ): void;
-  createExtractJob(
-    destination: File | File[] | string | string[],
-    callback: JobCallback
-  ): void;
+  createExtractJob(destination: string | File, callback: JobCallback): void;
   /**
    * Export model to Cloud Storage.
    *
    * @see [Jobs: insert API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/jobs/insert}
    *
-   * @param {string|string[]|File|File[]} destination Where the model should be exported
-   *    to. A string, string array, {@link
+   * @param {string|File} destination Where the model should be exported
+   *    to. A string or {@link
    *    https://cloud.google.com/nodejs/docs/reference/storage/latest/File File}
-   *    object, or File object array.
+   *    object.
    * @param {object} [options] - The configuration object.
-   * @param {string} [options.format="ML_TF_SAVED_MODEL"] - The format to export the data in. 
-   *    Allowed options are "ML_TF_SAVED_MODEL" or "ML_XGBOOST_BOOSTER". Default: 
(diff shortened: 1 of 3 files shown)


src/model.ts
```diff
@@ -15,7 +15,53 @@
...
+  Job,
+  Dataset,
+  ResourceCallback,
+  RequestCallback,
+  JobRequest,
+} from '.';
+import {JobMetadata} from './job';
+import bigquery from './types';
+
+// This is supposed to be a @google-cloud/storage `File` type. The storage npm
```
JustinBeckwith, line 32: Great comment, thank you for being this explicit.  I kinda think about this problem a lot.  Could I trouble you to file a top level tracking issue in https://github.com/googleapis/google-cloud-node on this problem?

system-test/bigquery.ts
```diff
@@ -625,10 +625,15 @@ describe('BigQuery', () => {
     });
   });
 
-  describe('BigQuery/Model', () => {
+  describe.only('BigQuery/Model', () => {
```
JustinBeckwith, line 628: Danger! We need to remove that :)

system-test/bigquery.ts
```diff
@@ -625,10 +625,15 @@ describe('BigQuery', () => {
...
     let model: Model;
+    const bucket = storage.bucket(generateName('bucket'));
+    const extractDest =
+      'gs://' + bucket.name + '/' + generateName('model-export');
+
+    before(async () => {
+      await bucket.create();
```
JustinBeckwith, line 635: I'm seeing a new bucket creation that happens every time, but no clean up.  Can we add an `after` hook here to remove the bucket we created?

test/model.ts
```diff
@@ -13,38 +13,108 @@
...
       '@google-cloud/common': {
         ServiceObject: FakeServiceObject,
+        util: fakeUtil,
       },
+      '@google-cloud/promisify': fakePfy,
     }).Model;
+
+    const modelCached = extend(true, {}, Model);
+
+    // Override all util methods, allowing them to be mocked. Overrides are
```
JustinBeckwith, line 82: Not really following what's happening here super closely, but generally I rely on `sinon` for doing this kinda thing

JustinBeckwith (commented):
Mostly looking good!

system-test/bigquery.ts
steffnay: There's a `[deleteBuckets](https://github.com/googleapis/nodejs-bigquery/blob/01f9fe1c55cc92db63164084fbe01d0a95b510f8/system-test/bigquery.ts#L105)` method to clean up all the buckets created by the system tests

src/model.ts
```diff
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2019 Google Inc. All Rights Reserved.
+ * Copyright 2020 Google Inc. All Rights Reserved.
```
JustinBeckwith, line 2: We're supposed to keep these dates to match the original introduction of the file, not the latest year.  We should keep this as is!

system-test/bigquery.ts
```diff
@@ -674,6 +679,17 @@ describe('BigQuery', () => {
       const [metadata] = await model.getMetadata();
       assert.strictEqual(metadata.friendlyName, friendlyName);
     });
+
+    it('should extract a model', () => {
```
JustinBeckwith, line 683: Let's make this an async function, and then use `await model.createExtractJob` to avoid the promise chaining.  As is, this test won't work correctly.

JustinBeckwith (approved):
LGTM with a few nits

PR #844, opened 2020-08-17
docs(samples): adds Create model sample

laljikanjareeya:
Adds Create Model Sample(bigquery_create_model)
Here is [Java sample reference](https://github.com/googleapis/java-bigquery/blob/ac8ed1904d9b53018e47b868b1a755e099ec66ee/samples/snippets/src/main/java/com/example/bigquery/CreateModel.java).


samples/createModel.js
```diff
@@ -0,0 +1,66 @@
...
+    // Run query to create a model
+    const [job] = await bigquery.createQueryJob(queryOptions);
+
+    // Wait for the query to finish
+    await job.getQueryResults();
+
+    console.log(`Model ${modelId} created.`);
+  }
+  // [END bigquery_create_model]
+  createModel();
```
steffnay, line 64: This needs to be inside the region tag

steffnay, line 64: This needs to be inside the region tag.

steffnay (approved):

PR #849, opened 2020-08-24
feat(constructor): add option to set baseUrl using environment variable

### LINKED ISSUE #568: Add option to use environment variable to set baseUrl for BigQuery constructor

steffnay:
Adds option to set `baseUrl` using `BIGQUERY_BASE_URL` environment variable.

Fixes #568 

🦕


test/bigquery.ts
```diff
@@ -213,6 +221,17 @@ describe('BigQuery', () => {
...
 
+    it('should allow overriding the baseUrl', () => {
+      const baseUrl = 'http://bigquery.test.url.com';
+      setHost(baseUrl);
+
+      bq = new BigQuery();
+      const calledWith = bq.calledWith_[0];
+      assert.strictEqual(calledWith.baseUrl, baseUrl);
+
+      unsetVariables();
```
stephenplusplus, line 232: It's possible this line won't be reached. We should cache the value of `process.env.BIGQUERY_EMULATOR_HOST` and restore it in an `after` hook.

stephenplusplus (changes requested):

test/bigquery.ts
```diff
@@ -170,6 +170,14 @@ describe('BigQuery', () => {
   });
 
   describe('instantiation', () => {
+    function setHost(host: string) {
```
stephenplusplus, line 173: In other test suites, we'll do a before/after to cache/restore "process.env.BIGQUERY_EMULATOR_HOST" to any value that may have been set by the user before running the tests.

stephenplusplus (changes requested):

src/bigquery.ts
```diff
@@ -251,9 +251,14 @@ export class BigQuery extends common.Service {
     options = Object.assign({}, options, {
       apiEndpoint: options.apiEndpoint || 'bigquery.googleapis.com',
     });
+
+    const baseUrl =
+      process.env.BIGQUERY_EMULATOR_HOST ||
+      `https://${options.apiEndpoint}/bigquery/v2`;
```
stephenplusplus, line 257: I've never been fully clear on the right way to go about this, and maybe each library is different-- should the env var replace everything, including the "/bigquery/v2" path? Or just the value of "options.apiEndpoint", etc.

steffnay: We want it to replace everything, which is also how it's done in [nodejs-storage](https://github.com/googleapis/nodejs-storage/blob/785b12f09919700c30106920ba3d4e578959591f/src/storage.ts#L403)

src/table.ts
```diff
@@ -1397,6 +1397,14 @@ class Table extends common.ServiceObject {
     }
 
     const dup = streamEvents(duplexify());
+    let uri: string;
+
+    if (this.bigQuery.baseUrl) {
+      const baseUrl = this.bigQuery.baseUrl.replace(/\/+$/, ''); // Remove trailing slashes
```
stephenplusplus, line 1403: Should this be done in the BigQuery constructor to avoid double-trailing slashes anywhere else as well? I think we probably have code somewhere upstream before a request goes out that would catch that, but it wouldn't hurt to clean up what we can when we get it.

src/bigquery.ts
stephenplusplus: We do a bit more there, too: https://github.com/googleapis/nodejs-storage/blob/d6dfdd244511d8b3ffaab2f561f272e8d81a2e8a/src/storage.ts#L431

- inserts a protocol if one is missing
- removes trailing slashes

It also treats the emulator env var as the `apiEndpoint` that gets passed to `super(config, ...)`, where this PR only uses it for `baseUrl`. Overall, it might be good to sync these two up, having this implementation match Storage.

steffnay: The reasons why they are divergent is that BigQuery expects that the provided `apiEndpoint` will not have a protocol and will not be manipulated before being set in the Service config. If changing the expected behavior for `apiEndpoint` is okay, I can go ahead and make it totally follow Storage.

steffnay: @stephenplusplus I went ahead and matched Storage's implementation. The major difference is that BigQuery assumed that apiEndpoint would be set on the client without a protocol. I'm thinking that changing this might be breaking to some if they are using this property for anything.

stephenplusplus: Thanks. It looks like the new code will handle when a user provides an `apiEndpoint` without a protocol when it passes through the `sanitizeEndpoint` function. So, I don't think there's a breaking path there, but let me know if I missed it.

test/table.ts
```diff
@@ -1698,6 +1698,39 @@ describe('BigQuery/Table', () => {
           })
           .emit('writing');
       });
+
+      describe('with BIGQUERY_EMULATOR_HOST baseUrl set', () => {
```
stephenplusplus, line 1702: I don't believe these tests should live here, as this is behavior that is performed in the `BigQuery` class.

src/bigquery.ts
```diff
@@ -376,6 +393,13 @@ export class BigQuery extends common.Service {
     });
   }
 
+  private static sanitizeEndpoint(url: string) {
```
stephenplusplus, line 396: Could you please add a `describe` block to test this function individually in test/bigquery.ts? You can probably just copy over the tests from Storage, now that we're modeled after those functions.

test/bigquery.ts
```diff
@@ -235,12 +244,12 @@ describe('BigQuery', () => {
       assert.deepStrictEqual(calledWith.scopes, expectedScopes);
     });
 
-    it('should not modify options argument', () => {
+    it('should not correctly modify options argument', () => {
```
stephenplusplus, line 247: Does this test still work as intended? It was making sure the object the user gave didn't have any properties tampered with by the constructor.

src/bigquery.ts
```diff
@@ -248,12 +250,27 @@ export class BigQuery extends common.Service {
...
+    let apiEndpoint = 'https://bigquery.googleapis.com';
+
+    const EMULATOR_HOST = process.env.BIGQUERY_EMULATOR_HOST;
+
+    if (typeof EMULATOR_HOST === 'string') {
+      apiEndpoint = BigQuery.sanitizeEndpoint(EMULATOR_HOST);
+    }
+
+    if (options.apiEndpoint) {
+      apiEndpoint = BigQuery.sanitizeEndpoint(options.apiEndpoint);
```
stephenplusplus, line 262: Maybe a new test to make sure `options.apiEndpoint` trumps the emulator variable.

stephenplusplus (changes requested):

test/bigquery.ts
steffnay: @stephenplusplus This is the issue I was referencing in a previous comment. The library originally ensured that no properties were mutated before being set, but if we are going to use baseUrl to set apiEndpoint, we have to then either assume the apiEndpoint will now be changed to include a protocol before being set (before it always expected the value to not include a protocol), or include a check for whether apiEndpoint has a protocol whenever we're using it to set a request uri in other places in the library. I Do you have an opinion on this?

steffnay: For example, [here](https://github.com/googleapis/nodejs-bigquery/blob/1e56383da5e6d8ce1335a711b32fea1155bddada/src/table.ts#L1417) is how we originally set the uri. If we are setting the apiEndpoint using provided baseUrl, I think we will have to assume apiEndpoint value includes the protocol.

stephenplusplus: Gotcha, thanks. This test was taking itself a bit more literally, I think, and making sure that the object itself wasn't changed by our code-- i.e., if we took their config object, overwrote their properties / added new ones, etc. We just want to basically leave the user's input object alone by making sure we're working from a cloned object in the constructor.

src/bigquery.ts
steffnay: @stephenplusplus What I mean is that if a user is used to initializing a client with `const client = new BigQuery({apiEndpoint: 'test.com'})` they would be used to `client.apiEndpoint` returning `'test.com'`, whereas now it will return `'https://test.com'`.

stephenplusplus: Does that affect anything, in terms of the address that's built to make API requests with? Users shouldn't be reading the `apiEndpoint` property, that's an internal/only-we-care thing.

steffnay: No, it doesn't affect anything. I was just concerned that I was missing a reason why reading this property might matter to users, but sounds like it's totally not a concern and we're all good.

stephenplusplus: It's great that you thought of that, I wouldn't have even considered that (...as was clear by my confusion). Sorry about that!

src/table.ts
```diff
@@ -1414,7 +1414,7 @@ class Table extends common.ServiceObject {
             },
           } as {},
           request: {
-            uri: `https://${this.bigQuery.apiEndpoint}/upload/bigquery/v2/projects/${this.bigQuery.projectId}/jobs`,
+            uri: `${this.bigQuery.apiEndpoint}/upload/bigquery/v2/projects/${this.bigQuery.projectId}/jobs`,
```
stephenplusplus, line 1417: It would probably be better if we changed the test, instead:

```diff
      it('should pass the correct request uri', done => {
        makeWritableStreamOverride = (
          stream: stream.Stream,
          options: MakeWritableStreamOptions
        ) => {
          const uri =
-           'https://bigquery.googleapis.com/upload/bigquery/v2/projects/' +
+           table.bigQuery.apiEndpoint + '/upload/bigquery/v2/projects/' +
            table.bigQuery.projectId +
            '/jobs';
          assert.strictEqual(options.request.uri, uri);
          done();
        };

        table.createWriteStream_().emit('writing');
      });
```

stephenplusplus (changes requested):

src/table.ts
stephenplusplus: I snuck a commit in to do this so we can merge and release in 5.3.0 along with another feature.

stephenplusplus (approved):

PR #864, opened 2020-09-28
fix: do not retry jobs.insert when it flakes

### LINKED ISSUE #857: Auto-retry job creation occasionally fails if job was created on first request
1) Is this a client library issue or a product issue?
It's a client library issue.

2) Did someone already solve this?
No, but someone did solve this in the associated Java client library:
- issue: https://github.com/googleapis/google-cloud-java/issues/2867
- fix: https://github.com/googleapis/google-cloud-java/pull/2887

3) Do you have a support contract?
No

#### Steps to reproduce

1. issue a job create request, leaving the library to choose a random ID, and leaving `autoRetry` on
2. occasionally the request will fail, and thus will be retried,
3. however the job will have actually been created in BigQuery, and the retry will fail with an error: `bigquery error: Already Exists: Job <id>`

The Java client library also had this issue, but was fixed in https://github.com/googleapis/google-cloud-java/pull/2887 by getting the relevant job if job creation failed due to this error (and if the job ID was randomly generated by the client library).

stephenplusplus:
Fixes #857

This is an attempt to handle a situation where we need to create an ID for a job. Before this PR:

- createJob() is called
- we generate a UUID and create a job
- the API processes the request and creates a job in the background
- the API call flakes and returns an error
- our auto-retry logic attempts the request again
- the API call falls with a 409: Already Exists error
- we return the 409 to the user

This PR will make it so that when the request is "auto-retried" and returns the 409, we'll ignore it, and instead refresh the metadata of the Job, treating that as if it was the original response from the create request.

- createJob() is called
- we generate a UUID and create a job
- the API processes the request and creates a job in the background
- the API call flakes and returns an error
- our auto-retry logic attempts the request again
- the API call falls with a 409: Already Exists error
- we ~~return the 409 to the user~~ refresh the metadata from the Job and return it to the user

#### To Dos

- [x] Unit tests


steffnay:
@stephenplusplus this looks good to me so far

steffnay (approved):

BenBirt:
Thanks for implementing this @stephenplusplus !

PR #872, opened 2020-10-12
fix: return error when custom getQueryResults() timeout has been exceeded 

### LINKED ISSUE #823: Timeout overriding does not works
Hi Team,

I want to override default time out value while making call to Big Query operation . I tried passing timeout value with options
`  const options = {
	timeoutMS :1
  }
` But I don't see any effect of this on method . I am not using Job's. Can some one please help me to understand why this is failing and how to override timeout value ?

steffnay:
It is currently possible to set `options.timeoutMs` when calling `Job.getQueryResults()`. The `timeoutMs` value is how long to wait for the query to complete, in milliseconds, before the request times out and returns. If the timeout is exceeded before the query completes, the API returns a response that includes `jobComplete: false`.  However, due to the way that auto-pagination is handled, the client still continues to make `getQueryResults()` requests until the query completes.

This change updates `Job.getQueryResults()` to return an error when a custom timeout has been exceeded, rather than continuing to check for query results.

`Job.createQueryJob()` is also updated to allow `job.configuration.jobTimeoutMs` to be set.

Fixes #823 🦕


src/job.ts
```diff
@@ -426,6 +428,15 @@ class Job extends Operation {
         let nextQuery: {} | null = null;
         if (resp.jobComplete === false) {
           // Query is still running.
+          // If timeout override was provided, return error.
+          if (timeoutOverride) {
+            const err = new Error(
+              `The query did not complete before ${timeoutOverride}ms`
+            );
+            callback!(err, null, null, resp);
```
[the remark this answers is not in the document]
steffnay: Good call.

src/job.ts
```diff
@@ -405,6 +405,8 @@ class Job extends Operation {
 
     delete qs.job;
 
+    const timeoutOverride = qs.timeoutMs ? qs.timeoutMs : false;
```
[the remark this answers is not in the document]
steffnay: Thanks!

src/bigquery.ts
```diff
@@ -1278,6 +1280,11 @@ export class BigQuery extends common.Service {
       },
     };
 
+    if (query.jobTimeoutMs) {
```
stephenplusplus, line 1283: Sorry I missed this one. If this can be 0, we have to check that it's defined as a number.

src/bigquery.ts
```diff
@@ -93,7 +93,7 @@ export type Query = JobRequest<bigquery.IJobConfigurationQuery> & {
   types?: string[] | string[][] | {[type: string]: string[]};
   job?: Job;
   maxResults?: number;
-  timeoutMs?: number;
```
stephenplusplus, line 96: Is this the breaking change part? Do you know the original intention of having `timeoutMs` here? I couldn't find that we handled `timeoutMs` ourselves, so maybe we thought that's what the API called it, so we were just saying that we accept it here, thinking the API would do something with it?

steffnay: Yeah, I was assuming removing this property would be considered breaking. And my theory is that `timeoutMs` ended up here by accident because it is part of a [QueryRequest](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#QueryRequest), but it was never actually being used for anything by the API since it isn't part of [JobConfigurationQuery](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfigurationQuery).

And I'm not totally clear on the exact boundaries of what is considered a breaking change in our libraries. Does the fact that job.getQueryResults() will now return an error if a user's timeout has been exceeded, rather than just continuing to check for results regardless of the value provided, count as breaking?

src/bigquery.ts
steffnay: Thanks for catching that!

src/bigquery.ts
stephenplusplus: > Yeah, I was assuming removing this property would be considered breaking.

TS-type breaks aren't too scary to me, because the developer will have a chance to catch it before anything is running in production.

>  Does the fact that job.getQueryResults() will now return an error if a user's timeout has been exceeded, rather than just continuing to check for results regardless of the value provided, count as breaking?

I would think this is more of a fix, in that now a user providing a timeout will actually get their expected results.

In both of these cases, the affected users should be a pretty low number. However, I still don't mind the 'better safe than sorry' approach, and releasing this under a major bump. ¯\_(ツ)_/¯

steffnay: @stephenplusplus I think you're right that it's actually more of a bug fix and we have a few users who need this corrected, so I think it's fine to go ahead and merge.

stephenplusplus (approved):

PR #873, opened 2020-10-12
feat: optionally wrap INT64 in BigQueryInt or provide a custom INT64 value type cast options

### LINKED ISSUE #6: losing precision in converting TIMESTAMP and INT64 to Javascript Number
_From @c0b on October 9, 2016 5:5_

https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1648#issuecomment-251383826

The BigQuery TIMESTAMP has up to microseconds precision, but when converting to a JavaScript Date, it becomes up to milliseconds

https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1648#issuecomment-251249300

A JavaScript Number is really only a FLOAT64, there is no real INT64, so during conversion some precision is lost:

``` js
$ node ./bigquery/queries.js sync 'SELECT ARRAY<INT64>[0x7fff1234deadbeef, -0x8000000000000000] AS example_array'
{ err: null,
  rows: [ { example_array: [ 9223110580161593000, -9223372036854776000 ] } ],
  nextQuery: null,
  apiResponse: 
   { kind: 'bigquery#queryResponse',
     schema: { fields: [ { name: 'example_array', type: 'INTEGER', mode: 'REPEATED' } ] },
     jobReference: { ... },
     totalRows: '1',
     rows: [ { f: [ { v: [ { v: '9223110580161593071' }, { v: '-9223372036854775808' } ] } ] } ],
     totalBytesProcessed: '0',
     jobComplete: true,
     cacheHit: false } }
Received 1 row(s)!
[ { example_array: [ 9223110580161593000, -9223372036854776000 ] } ]
```

I don't really have a solution, please suggest when application need this much precision

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#1681_

steffnay:
Adds ability to wrap 'INT64' values in a BigQueryInt type.

- Add BigQueryInt class
- Allow user to pass a custom typeCastFunction to be used to convert integer values
- Give user option to specify schema fields to be converted by typeCastFunction

Fixes #6  🦕


src/bigquery.ts
```diff
@@ -765,6 +792,46 @@ export class BigQuery extends common.Service {
...
+   * const {BigQuery} = require('@google-cloud/bigquery');
+   * const bigquery = new BigQuery();
+   *
+   * const largeIntegerValue = Number.MAX_SAFE_INTEGER + 1;
+   *
+   * customTypeCast = (value) => {
+   *    return value.split('');
+   * };
+   *
+   * const wrapIntegers = {
```
stephenplusplus, line 813:
```suggestion
   * const options = {
   *   integerTypeCastFunction: value => value.split(),
   * };
```

src/bigquery.ts
```diff
@@ -765,6 +792,46 @@ export class BigQuery extends common.Service {
...
+   *    integerTypeCastFunction: customTypeCast
+   * };
+   *
+   * const bqInteger = bigquery.int(largeIntegerValue, wrapIntegers);
+   *
+   * const customValue = bqInteger.valueOf();
```
stephenplusplus, line 819:
```suggestion
   * const customValue = bqInteger.valueOf();
   * // customValue is the value returned from your `integerTypeCastFunction`.
```

src/bigquery.ts
```diff
@@ -797,6 +864,31 @@ export class BigQuery extends common.Service {
...
+      throw new Error(
+        'We attempted to return all of the numeric values, but ' +
+          (value.schemaFieldName ? value.schemaFieldName + ' ' : '') +
+          'value ' +
+          value.integerValue +
+          " is out of bounds of 'Number.MAX_SAFE_INTEGER'.\n" +
+          "To prevent this error, please consider passing 'options.wrapNumbers' as\n" +
+          '{\n' +
+          '  integerTypeCastFunction: provide <your_custom_function>\n' +
+          '  properties: optionally specify property name(s) to be custom casted\n' +
```
stephenplusplus, line 885:
```suggestion
          '  fields: optionally specify field name(s) to be custom casted\n' +
```

?

src/bigquery.ts
```diff
@@ -1826,7 +1934,9 @@ export class BigQuery extends common.Service {
       query.job.getQueryResults(query, callback as QueryRowsCallback);
       return;
     }
-    this.query(query, {autoPaginate: false}, callback);
+    const wrapIntegers = query.wrapIntegers ? query.wrapIntegers : false;
```
stephenplusplus, line 1937: I think we can add an options object to this function to match how `query()` expects `wrapIntegers`.

src/bigquery.ts
stephenplusplus: This and the next are just some suggestions that could make it easier for the user to understand how the options all plug-in together.

stephenplusplus:
@steffnay What is the part that is a breaking change? It looks like an upgrade wouldn't break anything for a user, but let me know if I missed it.

steffnay:
@stephenplusplus it's not breaking, updated the PR title.

stephenplusplus (approved):

PR #877, opened 2020-10-19
fix: Detect Geography type during parameterized query

steffnay:
Adds support for detecting a Geography value's type during parameterized query.

🦕


JustinBeckwith (approved):

src/bigquery.ts
```diff
@@ -1007,7 +1009,11 @@ export class BigQuery extends common.Service {
     }
 
     function isCustomType({type}: ValueType): boolean {
-      return type!.indexOf('TIME') > -1 || type!.indexOf('DATE') > -1;
+      return (
```
stephenplusplus, line 1012: I don't believe this part is tested.

stephenplusplus (changes requested):

src/bigquery.ts
steffnay: updated.

stephenplusplus (approved):

PR #892, opened 2020-11-10
feat: Add support for Table IAM policies

### LINKED ISSUE #874: BigQuery: implement support for table ACLs
This FR is for tracking implementation support for table ACLs in the BigQuery manual library.

Specific implementation guidance has previously been shared out of band with maintainers on the necessary changes.

steffnay:
Adds the ability to get/set a table's IAM policy and test permissions.

Fixes #874 🦕


src/table.ts
```diff
@@ -2237,6 +2246,129 @@ class Table extends common.ServiceObject {
...
+    optionsOrCallback?: GetPolicyOptions,
+    cb?: PolicyCallback
+  ): void | Promise<PolicyResponse> {
+    const options =
+      typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+    const callback =
+      typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
+
+    if (
+      options.requestedPolicyVersion &&
```
JustinBeckwith, line 2268: Just to double check - `options.requestedPolicyVersion` will return falsey values for `null`, `undefined`, `0`, and empty strings.  That may be fine here, but I like to make sure that's the intent :)

JustinBeckwith (approved):

src/table.ts
steffnay: Just refactored this because I think it's better to just return the default version in those cases.

PR #895, opened 2020-11-14
docs(samples): Update table.getRows and dataset access samples

steffnay:
- Updates samples for query auto-pagination and paginating through table rows.

 🦕


samples/browseTable.js
```diff
@@ -0,0 +1,61 @@
...
+    const options = {
+      autoPaginate: false,
+      maxResults: 10,
+    };
+
+    // Retrieve rows.
+    table.getRows(options, manualPaginationCallback);
+  }
+  // [END bigquery_browse_table]
+  browseTable();
```
JustinBeckwith, line 59: This has to be inside of the region tag, or the code will never be invoked when the customer tries it.

samples/queryPagination.js
```diff
@@ -0,0 +1,72 @@
...
+    // Get rows.
+    const [rows] = await destinationTable.getRows(getRowsOptions);
+
+    console.log('Rows:');
+    rows.forEach(row => {
+      console.log(row);
+    });
+  }
+  // [END bigquery_query_pagination]
+  queryPagination();
```
JustinBeckwith, line 70: Same issue, needs to be inside the region tag

JustinBeckwith (changes requested):

samples/browseTable.js
```diff
@@ -0,0 +1,61 @@
...
+    // const datasetId = 'my_dataset'; // Existing dataset
+    // const tableId = 'my_table'; // Existing table
+
+    // Create table reference.
+    const dataset = bigquery.dataset(datasetId);
+    const table = dataset.table(tableId);
+
+    function manualPaginationCallback(err, rows, nextQuery) {
+      console.log('Rows:');
+      rows.forEach(row => console.log(row));
```
tswast, line 41: It'd be helpful to show what operations are possible on rows. Are they indexable by number or column name, for example.

(We don't actually do that in Python https://github.com/googleapis/python-bigquery/blob/9d43d2073dc88140ae69e6778551d140430e410d/samples/browse_table_data.py but if I did it again, I'd prefer listing rows from a sample table with a well-known schema.

samples/queryPagination.js
```diff
@@ -0,0 +1,72 @@
...
+    // Wait for job to complete.
+    await job.getQueryResults();
+
+    // For all options, see https://cloud.google.com/bigquery/docs/reference/v2/tabledata/list
+    const getRowsOptions = {
+      maxResults: 20,
+    };
+
+    // Get rows.
+    const [rows] = await destinationTable.getRows(getRowsOptions);
```
tswast, line 62: Is it possible to paginate over the results without a destination table?

samples/queryPagination.js
```diff
@@ -0,0 +1,72 @@
...
+    const queryOptions = {
+      query: query,
+      destination: destinationTable,
+    };
+
+    // Run the query as a job.
+    const [job] = await bigquery.createQueryJob(queryOptions);
+
+    // Wait for job to complete.
+    await job.getQueryResults();
```
tswast, line 54: I recall that the Node.js library used to download all result rows in one of these methods. Is that no longer the case?

samples/browseTable.js
steffnay: I think rewriting this sample to be able to do that will introduce complexity that decreases its overall quality. This library ultimately infers the table's project ID from the instance of the client, so creating a reference to a public datasets table in order to call getRows() would require showing users to manually reset the projectId, which would likely create a lot of confusion. Alternately, creating a table using a query would require calling job.getQueryResults() in order to determine when the job is finished creating the table, and would then be confusing because that method also returns the rows.

steffnay: Oh, meant to also add that row values are accessible in the same way that would be expected with objects in Node.js, so I think we'd be safe in opting to not explicitly show it here.

tswast: Hmmm... I don't like that it's not easier to download rows from across projects, but that shouldn't block this sample.

> job.getQueryResults() in order to determine when the job is finished creating the table, and would then be confusing because that method also returns the rows.

See my other comment. I have a concern if `job.getQueryResults()` downloads all rows. The main point of pagination is to save on memory resources. It's bad if it's not possible to do so with this client for query results. (Maybe there's a feature request hidden in here for a getQueryResultsStream method?)

samples/queryPagination.js
steffnay: Yeah, updated.

samples/browseTable.js
steffnay: You can paginate. I was just thinking that the reason for using `job.getQueryResults()` in this way might not be clear to users, but I think you're right that it's better to do it that way. Updated.

tswast (approved):

JustinBeckwith (approved):

tmatsuo:
Re: snippet-bot comment

Seems like it's going to delete a region tag which is used in devsite. It is ok to merge if you plan to update the devsite usage to the new location soon.

If you think it's a false positive, let me know.

tmatsuo:
However, the correct course of action from the Drift team's recommendation is to 1) first duplicate the region tag, 2) update devsite, then 3) remove the old region tag.

steffnay:
@tmatsuo My bad, will follow that process from now on! I'm waiting for devsite update to be reviewed, so should land soon.

PR #900, opened 2020-12-02
fix: update returned Job with API-determined location in getMetadata

### LINKED ISSUE #896: table.load() fails to find its own job when location is not specified in v5.x API
#### Environment details

  - OS: appengine-standard environment, local MacOS
  - Node.js version: 12.x, 14.x
  - npm version: 6.13.8
  - `@google-cloud/bigquery` version: 5.x (tested 5.0.0, 5.4.0). Works in v4.x

#### Steps to reproduce

  1. Create a dataset that is not in US (e.g. `australia-southeast-1`)
  2. Use the sample code to perform a load from https://github.com/googleapis/nodejs-bigquery/blob/master/samples/loadJSONFromGCS.js and omit the `location` option in `metadata`.
  3. The following error occurs

```
Error: Not found: Job my-test-project:396748ce-c4ec-449b-975d-1234567abb15
```

As mentioned this was working in v4. Not great if the api can internally create a job and lose track of it. This is not a huge priority because the workaround is to be explicit about specifying the `location`, however this is an undocumented change in v5 that may catch people unaware (I found root cause while documenting this issue).

laljikanjareeya:
`reqOpts` of  `getMetadata` has been set while initializing job object(constructor), once object initialized the there wasn't any impact of job's location changes in getMetadata's query string.

**Location is passing currently in query string:**
```ts
const job = new Job(bigquery, 'jobId', {
    location: 'australia-southeast1'
});

job.getMetadata((err, metadata, apiResponse) => { });
```

**query string is `undefined`**
```ts
const job = new Job(bigquery, 'jobId');
job.location = 'australia-southeast1';
job.getMetadata((err, metadata, apiResponse) => { });
```

Fixes #896


src/job.ts
```diff
@@ -190,41 +194,6 @@ class Job extends Operation {
...
-       * // If the callback is omitted, we'll return a Promise.
-       * //-
-       * job.getMetadata().then((data) => {
-       *   const metadata = data[0];
-       *   const apiResponse = data[1];
-       * });
-       */
-      getMetadata: {
-        reqOpts: {
-          qs: {location},
```
stephenplusplus, line 225: I wonder if a live getter would work here:

```js
getMetadata: {
  reqOpts: {
    qs: {
      get location() {
        return this.location;
      }
    }
  }
}
```

mbyrne00:
I’m not sure if this suits the style/structure of your tests or not but a high level test to create a job in Australia (or any other non-us) and wait for it to complete would highlight the exact use case nicely to the next developer. Whichever combination caused job not found. I see the tests seem quite targeted for specific functions so if this doesn’t suit or is too much hassle feel free to ignore.

src/job.ts
laljikanjareeya: @stephenplusplus The above given solution throws below Error becase of scoping issue:
 `RangeError: Maximum call stack size exceeded javascript`

I did try with below but it also doesn't work:

```ts
getMetadata: {
  reqOpts: {
    qs: () => {
      this.location;
    },
  },
},
```

stephenplusplus: Woops, I made a diff of what should work:

```diff
diff --git a/src/job.ts b/src/job.ts
index ab25289..e12befb 100644
--- a/src/job.ts
+++ b/src/job.ts
@@ -126,10 +126,7 @@ class Job extends Operation {
     options?: QueryResultsOptions
   ) => ResourceStream<RowMetadata>;
   constructor(bigQuery: BigQuery, id: string, options?: JobOptions) {
-    let location;
-    if (options && options.location) {
-      location = options.location;
-    }
+    let location: string;
 
     const methods = {
       /**
@@ -222,7 +219,12 @@ class Job extends Operation {
        */
       getMetadata: {
         reqOpts: {
-          qs: {location},
+          qs: {
+            get location() {
+              console.log({location})
+              return location;
+            },
+          },
         },
       },
     };
@@ -234,6 +236,15 @@ class Job extends Operation {
       methods,
     });
 
+    Object.defineProperty(this, 'location', {
+      get() {
+        return location;
+      },
+      set(_location) {
+        location = _location;
+      },
+    });
+
     this.bigQuery = bigQuery;
 
     if (options && options.location) {
```

laljikanjareeya: Done Thanks!!

stephenplusplus (approved):

PR #920, opened 2021-01-31
fix: handle null query parameter value

### LINKED ISSUE #915: Unable to pass null as a parameter value
Hi,

I am calling createQueryJob using this library ("@google-cloud/bigquery": "^5.5.0").  I have a parameter that is an array.  Originally my code was failing for empty arrays because I wasn't providing the type for the param.  So I added the type and all is good.  However, if I pass null as the value of the param, the code breaks.  I think the issue is with this line here:

https://github.com/googleapis/nodejs-bigquery/blob/master/src/bigquery.ts#L1107

Since value is null, value.type blows up.  I think it should perhaps be updated to if (value !== null && value.type)?

steffnay:
Updates `_getValue()` to check for `value`, which fixes the bug in handling `null` query parameter values.

Fixes #915 🦕


src/bigquery.ts
```diff
@@ -1104,7 +1104,7 @@ export class BigQuery extends common.Service {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private static _getValue(value: any, type: ValueType): any {
-    if (value.type!) type = value;
+    if (value && value.type) type = value;
```
stephenplusplus, line 1107: If it's possible for a user to provide a null value for a custom type, the line below would try to access `value.value` and throw. I think we may want to short-circuit `null` values instead of letting the function continue to execute:

```js
if (value === null) {
  return null;
}
```

steffnay: Makes sense. Updated.

stephenplusplus (approved):

test/bigquery.ts
```diff
@@ -1461,6 +1461,14 @@ describe('BigQuery', () => {
           geography.value
         );
       });
+
+      it('should handle null values', () => {
+        const value = null;
+        const type = 'TYPE';
+
+        sandbox.stub(BigQuery, '_isCustomType').returns(false);
```
stephenplusplus, line 1469:
```suggestion
```

stephenplusplus (changes requested):
Sorry, small thing.

stephenplusplus (approved):

PR #932, opened 2021-02-24
docs: update minor typo as per b/180010154

meredithslota:
Fixes #b/180010154 🦕

### DIFF
--- a/samples/updateViewQuery.js
+++ b/samples/updateViewQuery.js
@@ -44,7 +44,7 @@ function main(
     // Retrieve existing view metadata
     const [metadata] = await view.getMetadata();
 
-    // Uodate view query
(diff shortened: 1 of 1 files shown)


shollyman (approved):

JustinBeckwith:
@meredithslota minor nit - the prefix we use here (fix vs docs) has an impact on how the release tooling works.  When we see "fix", it signals to the release tooling that there's a change in the source code that requires shipping a new version of the package to package managers.  With "docs", we won't do that.  Since this is in sample code, I went ahead and switched it - hope that's ok!

stephenplusplus (approved):

meredithslota:
Oh hey thank you!

PR #953, opened 2021-05-01
fix: promise never returned on table.insert

### LINKED ISSUE #948: table.insert does not return a Promise
`table.insert` does not return a promise or has a dangling one

1) Is this a client library issue or a product issue?
Client library issue

2) Did someone already solve this?
Node

3) Do you have a support contract?
No

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - OS: MacOs
  - Node.js version: v14.16.1
  - npm version: 6.14.12
  - `@google-cloud/bigquery` version: 5.5.0

#### Steps to reproduce

1) Call `table.insert` without a callback and wrap it in a try/catch, as in:

```js
try {
  await this.table.insert({...});
} catch (error) {
 ...
}
```

2) If the data has a wrong type, then the call will have an Unhandled rejection.

Below are the errors:
```
Unhandled rejection
Unhandled rejection
PartialFailureError at Table._insert (/workspace/node_modules/@google-cloud/bigquery/build/src/table.js:1611:19)
                               at processTicksAndRejections (internal/process/task_queues.js:97:5)
                               at async Table._insertWithRetry (/workspace/node_modules/@google-cloud/bigquery/build/src/table.js:1548:24)
                               at async Table._insertAndCreateTable (/workspace/node_modules/@google-cloud/bigquery/build/src/table.js:1510:20)
```

larssn:

Fixes #948 

So the `table.insert` function never returned its promise. This PR fixes that.
It also gets rid of an unnecessary cast.

Two new tests were written:
- `it should return a promise if no callback is provided`
- `it should resolve to an array on success`

Code coverage goes down to 99.92, due to line 1916 being uncovered (even though thats what the new tests were for).

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -1904,14 +1904,17 @@ class Table extends common.ServiceObject {
         ? optionsOrCallback
         : ({} as InsertRowsOptions);
     const callback =
-      typeof optionsOrCallback === 'function'
-        ? optionsOrCallback
-        : (cb as InsertRowsCallback);
+      typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
 
-    this._insertAndCreateTable(rows, options).then(
-      resp => callback(null, resp),
-      err => callback(err, null)
-    );
+    const promise = this._insertAndCreateTable(rows, options);
+    if (callback) {
+      promise.then(
+        resp => callback(null, resp),
+        err => callback(err, null)
+      );
+    } else {
+      return promise.then(r => [r]);
+    }
   }
 
   /**
--- a/test/table.ts
+++ b/test/table.ts
@@ -2274,6 +2274,16 @@ describe('BigQuery/Table', () => {
       );
     });
 
+    it('should return a promise if no callback is provided', () => {
+      const promise = table.insert(data);
+      assert(promise instanceof Promise);
+    });
+
+    it('should resolve to an array on success', async () => {
+      const resp = await table.insert(data);
+      assert(Array.isArray(resp));
+    });
+
     it('should generate insertId', async () => {
       await table.insert([data[0]]);
       assert(


src/table.ts
```diff
@@ -1904,14 +1904,17 @@ class Table extends common.ServiceObject {
...
-      err => callback(err, null)
-    );
+    const promise = this._insertAndCreateTable(rows, options);
+    if (callback) {
+      promise.then(
+        resp => callback(null, resp),
+        err => callback(err, null)
+      );
+    } else {
+      return promise.then(r => [r]);
```
feywind, line 1916: It's sort of odd to me that the Promise returns an array, but the callback isn't called with one. This seems to maintain the method definition though.

feywind (approved):
I disagree with codecov, I think those lines are covered.

src/table.ts
```diff
@@ -1904,14 +1904,17 @@ class Table extends common.ServiceObject {
...
-        ? optionsOrCallback
-        : (cb as InsertRowsCallback);
+      typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
 
-    this._insertAndCreateTable(rows, options).then(
-      resp => callback(null, resp),
-      err => callback(err, null)
-    );
+    const promise = this._insertAndCreateTable(rows, options);
+    if (callback) {
```
sk-, line 1910: I don't think this is needed, as there's already a promisified version being returned. See https://github.com/googleapis/nodejs-bigquery/blob/a138347855f74d4e5c889dc42b00992c4a3808a6/src/table.ts#L2388.

I think the unhandled rejection may come from further down.

src/table.ts
larssn: Or it could be typescript having trouble determining the return type.
I'm not sure if there's a test for the callback version's return type.

src/table.ts
larssn: You're right, this PR fixes nothing then. Easy to miss. Not the biggest fan of code like that tbh.

cc @feywind

PR #984, opened 2021-08-08
docs: Update type annotations

steffnay:
Updates type annotations.

 🦕


src/bigquery.ts
```diff
@@ -1139,6 +1146,12 @@ export class BigQuery extends common.Service {
     callback: DatasetCallback
   ): void;
   createDataset(id: string, callback: DatasetCallback): void;
+  /**
+   * @callback DatasetCallback
+   * @param {?Error} err Request error, if any.
+   * @param {object} dataset The [dataset resource]{@link https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#resource}.
+   * @param {object} apiResponse The full API response.
+   */
```
shollyman, line 1154: I realize this is probably just a node-ism, but this idea of comment blocks one after another that aren't anchored to something outside the comment block weirds me out. :)

src/bigquery.ts
```diff
@@ -1690,6 +1712,15 @@ export class BigQuery extends common.Service {
   getJobs(options?: GetJobsOptions): Promise<GetJobsResponse>;
   getJobs(options: GetJobsOptions, callback: GetJobsCallback): void;
   getJobs(callback: GetJobsCallback): void;
+  /**
+   * @callback GetJobsCallback
+   * @param {?Error} err Request error, if any.
+   * @param {object} jobs An array of [Job]{@link https://cloud.google.com/bigquery/docs/reference/v2/jobs} objects.
```
shollyman, line 1718: Is this another node-ism, that the callback args apparently have key names but the response objects use integers for argument ordering?

shollyman (approved):
A couple of comments in the "I don't understand node and I'm not afraid to show it" vein.

bcoe:
@steffnay friendly note that I think you'll need to reopen this against `main`, GitHub's migration is supposed to do this for you, but I believe we bumped into an issue with their API this morning during the migration.

src/bigquery.ts
steffnay: I think that it is a TypeScript-ism because it looks like the pattern occurs in some samples [here](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#typedef-callback-and-param).

src/bigquery.ts
steffnay: Yeah, that's correct. It difference looks like this:
```
bigquery.getJobs(function(err, jobs) {
 if (!err) {
   // jobs is an array of Job objects.
  }
});

bigquery.getJobs().then(function(data) {
   const jobs = data[0];
});
```

PR #999, opened 2021-08-28
fix: update queryAsStream_ to set options with Query type

### LINKED ISSUE #947: Is it possible to get numbers as strings when streaming results?
If you run the following query:

```
const stream = client.createQueryStream({ query: "select 9999999999999999" });
```

and then stream the results with:

```
for await (const row of stream) {
  console.log({ row });
}
```

you will see:

```
{ row: { f0_: 10000000000000000 } }
```

I see something similar is mentioned in https://github.com/googleapis/nodejs-bigquery/issues/6, where it is claimed that this is not an issue because you can access the raw API response (e.g. of `job.getQueryResults`?). What I'm not sure of is if it is possible to _stream_ the raw API response?

steffnay:
This ensures query options that are provided within the single `Query` type parameter are properly set, which will allow the `wrapIntegers` option to work with streaming.

Fixes #947 🦕


JustinBeckwith (commented):
This generally looks ok due to the `extend` below, but it doesn't really match the pattern elsewhere in the library.  I'd love to see 2 things here:
1. a comment above line 2011 that explains why this is different
2. a test that validates the fix you're making so it doesn't regress

JustinBeckwith (approved):

PR #1000, opened 2021-08-28
fix(types): add string type to Table.query & Dataset.query

### LINKED ISSUE #942: Missing string type on Table.query & Dataset.query ?
#### Description

Calling `query` method on `Table` class using a 'string' --> type error. 
Work around is to simply wrap in a `Query` object but I think the types could be updated to support passing in a string - as is possible for the query method on `BigQuery` class.

#### Environment details

  - `@google-cloud/bigquery` version: 5.5.0

#### Current types

(for simplicity only including types without the callback option)

```js
// BigQuery https://github.com/googleapis/nodejs-bigquery/blob/master/src/bigquery.ts#L1807
query(query: Query, options?: QueryOptions): Promise<SimpleQueryRowsResponse>;
query(query: string, options?: QueryOptions): Promise<QueryRowsResponse>;

// Dataset https://github.com/googleapis/nodejs-bigquery/blob/master/src/dataset.ts#L1010
query(options: Query): Promise<QueryRowsResponse>;
// missing options: string ?

// Table https://github.com/googleapis/nodejs-bigquery/blob/master/src/table.ts#L2175
query(query: Query): Promise<SimpleQueryRowsResponse>
// missing query: string ?
```

#### Reasoning

The `Table.query` method calls `Dataset.query` which calls `BigQuery.query`. 
The `Dataset.query` method and `BigQuery.query` method suggest that they could handle string query arguments.

[Dateset.query](https://github.com/googleapis/nodejs-bigquery/blob/master/src/dataset.ts#L1016)
```js
if (typeof options === 'string') {
      options = {
        query: options,
      };
    }
```

[BigQuery.query](https://github.com/googleapis/nodejs-bigquery/blob/master/src/bigquery.ts#L1954)
```js
      if (typeof query === 'object' && query.dryRun) {
        (callback as SimpleQueryRowsCallback)(null, [], resp);
        return;
      }
```

[Table.query docs](https://googleapis.dev/nodejs/bigquery/latest/Table.html#query) redirect you to the [BigQuery.query docs](https://googleapis.dev/nodejs/bigquery/latest/BigQuery.html#query) which suggests the types should be the same.

#### Steps to reproduce

```js
import {BigQuery} from '@google-cloud/bigquery';

const queryStr = `SELECT corpus
              FROM \`bigquery-public-data.samples.shakespeare\` 
              LIMIT 10`,
const bqClient = new BigQuery();
bqClient.dataset('id').query(queryStr); // Type 'string' has no properties in common with type 'Query'.
bqClient
  .dataset('id')
  .table('table')
  .query(queryStr); // Type 'string' has no properties in common with type 'Query'.
```

#### Suggestions

Something like... update the `Table.query` and `Dataset.query` to accept `string`:

```js
// dataset.ts
  query(options: string): Promise<QueryRowsResponse>; // add this string type
  query(options: string, callback: SimpleQueryRowsCallback): void // add this string type
query(
    options: Query | string, // update to union type
    callback?: SimpleQueryRowsCallback
  ): void | Promise<QueryRowsResponse> {

//table.ts
  query(query: Query): Promise<SimpleQueryRowsResponse>;
  query(query: Query, callback: SimpleQueryRowsCallback): void;
  query(query: string): Promise<SimpleQueryRowsResponse>; // add this string type
  query(query: string, callback: SimpleQueryRowsCallback): void // add this string type
  query(
    query: Query | string, // update to union type
    callback?: SimpleQueryRowsCallback
  ): void | Promise<SimpleQueryRowsResponse> {
    if (typeof query === 'string') { // add in this check which is already in Dataset and BigQuery
      query = {
        query,
      };
    }
    this.dataset.query(query, callback!);
  }

```

Thanks for looking.

steffnay:

Fixes #942 🦕

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -1139,7 +1139,9 @@ class Dataset extends ServiceObject {
   }
 
   query(options: Query): Promise<QueryRowsResponse>;
+  query(options: string): Promise<QueryRowsResponse>;
   query(options: Query, callback: SimpleQueryRowsCallback): void;
+  query(options: string, callback: SimpleQueryRowsCallback): void;
   /**
    * Run a query scoped to your dataset.
    *
@@ -1151,7 +1153,7 @@ class Dataset extends ServiceObject {
    * @returns {Promise<QueryRowsResponse>} See {@link BigQuery#query} for full documentation of this method.
    */
   query(
-    options: Query,
+    options: Query | string,
     callback?: SimpleQueryRowsCallback
   ): void | Promise<QueryRowsResponse> {
     if (typeof options === 'string') {
--- a/src/table.ts
+++ b/src/table.ts
@@ -2271,6 +2271,7 @@ class Table extends common.ServiceObject {
   }
 
   query(query: Query): Promise<SimpleQueryRowsResponse>;
+  query(query: string): Promise<SimpleQueryRowsResponse>;
   query(query: Query, callback: SimpleQueryRowsCallback): void;
   /**
    * Run a query scoped to your dataset.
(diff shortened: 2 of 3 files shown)


JustinBeckwith (approved):
Looks great, but maybe make it a `fix(types):` so it shows up in the changelog for our TypeScript friends :)

PR #1023, opened 2021-10-08
feat: Add jobs.delete 

### LINKED ISSUE #1010: BigQuery - support jobs.delete
Feature request is to support the `jobs.delete` method in the BigQuery library.  Additional information about this feature has been shared out of band with library maintainers.

steffnay:

Fixes #1010 🦕


JustinBeckwith (dismissed):

src/job.ts
```diff
@@ -46,6 +46,8 @@ export type JobOptions = JobRequest<JobMetadata>;
 export type CancelCallback = RequestCallback<bigquery.IJobCancelResponse>;
 export type CancelResponse = [bigquery.IJobCancelResponse];
 
+export type DeleteCallback = RequestCallback<object>;
```
JustinBeckwith, line 49: `<object>` is super generic - what actually gets returned by the API?

JustinBeckwith: Maybe we should be extending this type?  https://github.com/googleapis/nodejs-common/blob/b1ceb532ac7bf4852f430c83af5a71618ed42f54/src/service-object.ts#L119

src/job.ts
```diff
@@ -370,6 +372,54 @@ class Job extends Operation {
     );
   }
 
+  delete(): Promise<any>;
```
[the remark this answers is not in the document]
JustinBeckwith: Wouldn't it be this type though?  
https://github.com/googleapis/nodejs-common/blob/b1ceb532ac7bf4852f430c83af5a71618ed42f54/src/service-object.ts#L275

I just want us to be as specific as possible.

steffnay:
@JustinBeckwith I realized I needed to refactor, so the type issue are no longer a thing.

bcoe (approved):

system-test/bigquery.ts
```diff
@@ -646,6 +646,25 @@ describe('BigQuery', () => {
...
+      const opts = {
+        configuration: {
+          query: {
+            query: 'SELECT 100 as foo',
+          },
+        },
+        location: 'us-east1',
+      };
+
+      const [job] = await bigquery.createJob(opts);
```
bcoe, line 660: It eoulf likely be worth adding some logic to cleanup stale jobs, kif this test were to fail mid-way we could leak some resources.

Is it possible to give the job an identifier, so that we could just list all jobs in a cleanup step, and delete any older than a certain age?

steffnay: for sure

PR #1040, opened 2021-11-05
fix(cloud-rad): move comments for TSDoc

fhinkel:
As we move our ref docs to cloud.google.com, we rely on TSDoc rather JSDoc.

TSDoc expects comments before the first overloaded function, we
currently have those on the last function. Those comments don't make
it into the d.ts files. We need to move comments to the first
overloaded function rather than the last one.

Internally b/190631834

Script used: https://github.com/fhinkel/cloud-rad-script/blob/main/moveComments.js

### DIFF
--- a/src/bigquery.ts
+++ b/src/bigquery.ts
@@ -1172,16 +1172,6 @@ export class BigQuery extends common.Service {
     );
   }
 
-  createDataset(
-    id: string,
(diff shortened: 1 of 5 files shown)


alexander-fenster (approved):
I think `fix` might fit better than `chore` here.

PR #1047, opened 2021-11-24
docs: change properties with function value to methods

fhinkel:
See https://github.com/microsoft/rushstack/issues/3031
    
Internally, see b/206943288

Drive-by fix: 
- chore: change import format for better docs
- Add description to BigQueryOptions properties

### DIFF
--- a/src/bigquery.ts
+++ b/src/bigquery.ts
@@ -286,9 +286,20 @@ export const PROTOCOL_REGEX = /^(\w*):\/\//;
 export class BigQuery extends Service {
   location?: string;
 
-  createQueryStream: (options?: Query | string) => ResourceStream<RowMetadata>;
-  getDatasetsStream: (options?: GetDatasetsOptions) => ResourceStream<Dataset>;
-  getJobsStream: (options?: GetJobsOptions) => ResourceStream<Job>;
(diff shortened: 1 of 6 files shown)


src/bigquery.ts
```diff
@@ -878,23 +892,7 @@ export class BigQuery extends common.Service {
    * A geography value represents a surface area on the Earth
    * in Well-known Text (WKT) format.
    *
-   * @method BigQuery.geography
```
tswast, line 881: I assume this was removed because it's documented at the actual method definition now?

src/bigquery.ts
```diff
@@ -1518,7 +1514,6 @@ export class BigQuery extends common.Service {
    * @param {?error} callback.err An error returned while making this request.
    * @param {Job} callback.job The newly created job.
    * @param {object} callback.apiResponse The full API response.
-   * @returns {Promise<JobResponse>}
```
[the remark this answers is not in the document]
fhinkel: Exactly.

src/bigquery.ts
fhinkel: Yes, it's redundant with TSDoc.

steffnay (approved):

PR #1098, opened 2022-04-13
chore: update dtsd version and discovery types

### LINKED ISSUE #1089: `jobs.IListParams.stateFilter` has wrong type
`stateFilter` accepts an array (and it's documented that way) but the type definition does not account for this, it only expects a single string.

I am specifically talking about this line: https://github.com/googleapis/nodejs-bigquery/blob/16d2a505400564cb49210d25f52fccfdaef42dd5/src/types.d.ts#L3568

Also here is the documentation that says it accepts an array: https://github.com/googleapis/nodejs-bigquery/blob/701ead4b3611aa31135baa5dfe9f7b7423ad4f2f/src/bigquery.ts#L1710

Right now I can workaround this by passing the array with `as any`.

steffnay:
Updates the version of discovery-tsd and regenerates types from the Discovery Doc.

Fixes #1089 🦕

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -570,6 +570,17 @@ declare namespace bigquery {
     count?: string;
   };
 
+  type ICloneDefinition = {
+    /**
+     * [Required] Reference describing the ID of the table that was cloned.
+     */
+    baseTableReference?: ITableReference;
+    /**
+     * [Required] The time at which the base table was cloned. This value is reported in the JSON response using RFC3339 format.
+     */
+    cloneTime?: string;
+  };
+
   /**
    * Message containing the information about one cluster.
    */
@@ -696,6 +707,10 @@ declare namespace bigquery {
      */
     evaluationTable?: ITableReference;
     /**
+     * Table reference of the test data after split.
+     */
+    testTable?: ITableReference;
+    /**
      * Table reference of the training data after split.
      */
     trainingTable?: ITableReference;
@@ -801,6 +816,10 @@ declare namespace bigquery {
      */
     location?: string;
     /**
+     * [Optional] Number of hours for the max time travel for all tables in the dataset.
+     */
+    maxTimeTravelHours?: string;
+    /**
      * [Output-only] Reserved for future use.
      */
     satisfiesPZS?: boolean;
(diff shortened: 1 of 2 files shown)


src/types.d.ts
```diff
@@ -808,19 +827,27 @@ declare namespace bigquery {
      * [Output-only] A URL that can be used to access the resource again. You can use this URL in Get or Update requests to the resource.
      */
     selfLink?: string;
+    /**
```
feywind, line 830: Are these generated? It might be nice to break out the {} here if not.

src/types.d.ts
```diff
@@ -2855,7 +3226,7 @@ declare namespace bigquery {
   };
 
   /**
-   * The type of a variable, e.g., a function argument. Examples: INT64: {type_kind="INT64"} ARRAY: {type_kind="ARRAY", array_element_type="STRING"} STRUCT>: {type_kind="STRUCT", struct_type={fields=[ {name="x", type={type_kind="STRING"}}, {name="y", type={type_kind="ARRAY", array_element_type="DATE"}} ]}}
+   * The data type of a variable such as a function argument. Examples include: * INT64: `{"typeKind": "INT64"}` * ARRAY: { "typeKind": "ARRAY", "arrayElementType": {"typeKind": "STRING"} } * STRUCT>: { "typeKind": "STRUCT", "structType": { "fields": [ { "name": "x", "type": {"typeKind: "STRING"} }, { "name": "y", "type": { "typeKind": "ARRAY", "arrayElementType": {"typekind": "DATE"} } } ] } }
```
feywind, line 3229: Seems like something happened to the formatting here? It looks like it was meant to be multi-line.

src/types.d.ts
```diff
@@ -3843,7 +4322,7 @@ declare namespace bigquery {
       /**
        * Filter for job state
        */
-      stateFilter?: 'done' | 'pending' | 'running';
+      stateFilter?: Array<'done' | 'pending' | 'running'>;
```
feywind, line 4325: Might be another nice place to break out types, though again if it's generated, probably not much to do...

feywind (approved):
Just a few suggestions!

Also it seems like the check annotation warnings are piling up, I don't think you should try to fix them in this PR, but maybe it's something to consider later if there's a cleanup pass.

shollyman (approved):

src/types.d.ts
steffnay: Yeah, they're generated.

PR #1178, opened 2023-01-17
fix(stream): Fix table#createReadStream with startIndex

### LINKED ISSUE #1099: `Job#getQueryResultsStream` fails when passed a `startIndex`
#### Environment details

  - OS:
  - Node.js version:
  - npm version:
  - `@google-cloud/bigquery` version: latest

#### Steps to reproduce

  1. Pass `startIndex` to `Job#getQueryResultsStream`
  2. First request to populate the stream is successful and starts at the correct index. 
  3. Second request fails with `When using a page token, you cannot specify an arbitrary startIndex`

It seems this is simply because `startIndex` should be used on the first fetch, and then discarded once page tokens can be used for the next query.

maximebedard:
Identical PR to https://github.com/googleapis/nodejs-bigquery/pull/1100, but for table#createReadStream with a startIndex.

Fixes #1099 🦕


loferris (changes requested):
This looks promising! We're going to need unit tests and integration tests before moving forward. It may be a while before I can write them myself, but if you'd like to take a look, I'd start by adding a unit test around here: https://github.com/googleapis/nodejs-bigquery/blob/df1dcd3899dd10f20988b938b31ac571c997bffd/test/table.ts#L1969

alvarowolfx:
@maximebedard do still have plans to write tests for this ? I can take this over.

PR #1258, opened 2023-08-11
feat: partial metadata projection

loferris:

This PR updates the documentation to specify how to get a partial metadata projection via the `table.get()` method, and clarified the difference between this method and `table.getMetadata()` in terms of being selective about metadata retrieved. A system test for this usage of `table.get()` has been added, as well as an expansion of the `table.getMetadata()` system test to help illustrate the differences between these two methods. This PR reflects an update to the BigQuery API across all languages, not a GH issue for this repo.


system-test/bigquery.ts
```diff
@@ -888,9 +887,25 @@ describe('BigQuery', () => {
       const description = 'catsandstuff';
       await table.setMetadata({description});
       const [metadata] = await table.getMetadata();
+      const metadataProps = Object.values(metadata);
+      assert.strictEqual(metadataProps.length, 18);
```
alvarowolfx, line 891: I think checking for the number of attributes here can lead the test to be flaky in the long term, as every time that the service team adds a new field, the test is going to break.

system-test/bigquery.ts
```diff
@@ -888,9 +887,25 @@ describe('BigQuery', () => {
...
+    it('should set & get partial metadata', async () => {
+      const options = {
+        view: 'BASIC',
+      };
+      const [basicMetadata] = await table.get(options);
+      const basicMetadataProps = Object.values(
+        Object.keys(basicMetadata.metadata)
+      );
+
+      assert.strictEqual(basicMetadataProps.length, 10);
```
alvarowolfx, line 904: Same comment around testing for the amount of fields

alvarowolfx (changes requested):
Checking the implementation guide, I think a good idea to test the different between getting a basic view vs full view is to check for the `lastModifiedTime` and/or `numBytes`, as in the basic view, those attributes are not computed by the backend.

https://github.com/googleapis/google-cloud-go/pull/6186/files#diff-62ca20d473019007434974e167cb4ad4be57a9f5b8b4ba834dfcf7abe3ee4f76

system-test/bigquery.ts
loferris: Good point!

loferris:
> Checking the implementation guide, I think a good idea to test the different between getting a basic view vs full view is to check for the `lastModifiedTime` and/or `numBytes`, as in the basic view, those attributes are not computed by the backend.
> 
> https://github.com/googleapis/google-cloud-go/pull/6186/files#diff-62ca20d473019007434974e167cb4ad4be57a9f5b8b4ba834dfcf7abe3ee4f76

I'm already checking against `numBytes` in the new system test so I can add `lastModifiedTime` as well.

alvarowolfx (approved):

PR #1263, opened 2023-08-16
fix: stricter typing for ServiceObject extensions

### LINKED ISSUE #1206: Typescript extending ServiceObject doesn't pass T causing service object methods to return untyped.
Various places that extend ServiceObject https://github.com/googleapis/nodejs-bigquery/blob/main/src/table.ts#L218 does this:

```ts
class Table extends ServiceObject {}
```

It should be

```ts
class Table extends ServiceObject<Table> {}
```

Because right now `dataset.table('foo').get()` isn't typed.

Typescript example of the issue: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMrQG7CQeQEYBWECYAPACrIC8ycIAngHzIDeAUMp8gOYRgAUASgBcycmwC+AbjZtQkWIhTk4eADYoIAD0ggAJgGc0mbBHxESraXPDR4SZADFgWiHpXrNOiPqPooWLiExGTOru6qGsws1mwIAPYgBmDIMPHx1Fa0Rh4aMmnxAHS8YBzIAPTlyAB6APyyCUkpeHBQmTHZTi5uuRAyLVDFfGWVNfVAA

loferris:

Fixes #1206 🦕


alvarowolfx (approved):

samples/getTable.js
```diff
@@ -37,6 +37,11 @@ function main(datasetId = 'my_dataset', tableId = 'my_table') {
     // Retrieve table reference
     const dataset = bigquery.dataset(datasetId);
     const [table] = await dataset.table(tableId).get();
+    /**Typescript Note:
+     * To narrow type from "any", cast the result of this method.
+     * const [table]: Dataset = await dataset.table(tableId).get();
+     * const table: Promise<GetResponse<Table>> = dataset.table(tableId).get();
```
ceifa, line 43: GetResponse is not exported by this library :/

PR #1271, opened 2023-09-13
feat: re-exporting types for google-common

### LINKED ISSUE #1268: Re-export ApiError from google-cloud/common to ease usage downstream

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

In our code, we want to check if an error that's returned by the BQ library is an `ApiError`, and if so, we want to re-use properties off it.

Within our code, we have something like this for getting a user's datasets:

```ts
import { BigQuery } from '@google-cloud/bigquery';
import { ApiError } from '@google-cloud/common';

client = new BigQuery({ /* options */ });

// ....

client
  .getDatasets()
  .then(([datasets]) => {
    // do something with the datasets
  })
  .catch((e) => {
    if (e instanceof ApiError) {
      if (e.code === 403 || e.code === 404) {
        // show user permission error
      }
    }
    // show user generic error message
  });
```

where to make the `instanceof ApiError` check work, we were adding `@google-cloud/common` to our dependencies and importing from there. We ended up with a bug where `@google-cloud/bigquery` got updated and so the version of `@google-cloud/common` in our dependencies no longer matched what `@google-cloud/bigquery` used, and the `instanceof ApiError` check now failed.

It'd be nice if `@google-cloud/bigquery` re-exported `ApiError` so that the above code snippet could be rewritten as :

```ts
import { ApiError, BigQuery } from '@google-cloud/bigquery';

client = new BigQuery({ /* options */ });

// ....

client
  .getDatasets()
  .then(([datasets]) => {
    // do something with the datasets
  })
  .catch((e) => {
    if (e instanceof ApiError) {
      if (e.code === 403 || e.code === 404) {
        // show user permission error
      }
    }
    // show user generic error message
  });
```

and then the `instanceof` check will never fail.

We could remove `@google-cloud/common` from our dependency list, and rely on it being available implicitly via how `node_modules` works, but that seems risky, and we could end up with being broken still depending on how yarn decides to hoist things. We've also looked to use duck typing, but that gives us less type strength.

 **Additional context**

loferris:

Fixes #1268  🦕

This PR introduces a re-exporting pattern to the library that allows for users to import `google-common` along with `bigquery`. This option ensures that users' version of `google-common` is the same as the one used in the BigQuery library.  This is also a way to gauge the value of re-exporting dependencies in the library.


alvarowolfx (approved):
Make sense to me. To we know if other libraries have that same kind of pattern ?

loferris:
> Make sense to me. To we know if other libraries have that same kind of pattern ?

I'm pretty sure the other languages do not, but it is a pattern in TypeScript libraries in general. I think I'll merge for now and see if we can work more with adding more targeted overloads later.

PR #1295, opened 2023-10-23
feat: implementing mocha retries

### LINKED ISSUE #1299: Add mocha retries to testing suite for samples

Due to transient service issues, CI tests for samples in this repo are flaky, which creates issues that are not actionable.
I think we should leverage our current testing framework and add mocha retries at the level of each test suite.
I considered writing our own retry logic, as well as adding retries to all tests that are async functions. I also contemplated changing the testing framework.
**Additional context**
This issue accompanies a PR that adds mocha retries to the sample tests.

loferris:

Fixes #1299 🦕


samples/test/authViewTutorial.test.js
```diff
@@ -36,7 +36,8 @@ const sharedViewId = generateUuid();
 
 const bigquery = new BigQuery();
 
-describe('Authorized View Tutorial', () => {
+describe('Authorized View Tutorial', function () {
+  this.retries(3);
```
alvarowolfx, line 40: I think doing this way, it will retry the whole `describe` section, which can be quite expensive to re run. I ran some tests here and seems like we can retry on a per spec basis.

```
describe('Authorized View Tutorial', () => { // Back to arrow function
  beforeEach(function(){
    console.log("this:", this.currentTest.title)
    this.currentTest.retries(2); // will try 3 times - n+1    
  })
```

Then I changed one of the tests to fail, to check for the retry behavior:

```
  it('should query stackoverflow', async () => {
    const output = execSync('node queryStackOverflow.js');
    assert.match(output, /Query Results:/);
    assert.match(output, /views/);
    assert.match(0,1)
  }); // Force failure
  ```

File: samples/test/datasets.test.js [hunk 2]
```diff
@@ -29,7 +29,8 @@ const datasetId = `${GCLOUD_TESTS_PREFIX}_datasets_${uuid.v4()}`.replace(
 
 const bigquery = new BigQuery();
 
-describe('Datasets', () => {
+describe('Datasets', function () {
+  this.retries(3);
```
alvarowolfx, line 33: If we want to try 3 times, we actually need to pass `2` as parameter, as the test will run once + number of retries.

alvarowolfx (changes requested):
Added a comment on how to make retries more granular and efficient.

alvarowolfx (approved):

PR #1351, opened 2024-04-02
fix: change startIndex option type from string to number

### LINKED ISSUE #1350: GetRowsOptions Type Error for startIndex: Expects a string when it should expect a number
#### Environment details

  - OS: macOS 14.4.1
  - Node.js version: 20.11
  - npm version: 10.2.4
  - `@google-cloud/bigquery` version: 7.5.2

#### Steps to reproduce

  1. Try to pass `startIndex` as an integer option to `getRows`. eg.`table.getRows({ startIndex: 31241 })`.
  2. Observe typescript error. Expects a string when this should be a number.
  3. Passing a number in as "string" allows it to compile.

#### Expected behaviour
`startIndex` should work with an integer. As it is right now I need to convert the integer to a string eg. `31241.toString()`

#### Relevant fixes in the code
https://github.com/googleapis/nodejs-bigquery/blob/74aa1501452c36af7969bb4a46b996485d9ca91b/src/types.d.ts#L5903
https://github.com/googleapis/nodejs-bigquery/blob/74aa1501452c36af7969bb4a46b996485d9ca91b/src/types.d.ts#L6055

### LINKED ISSUE #7230: GetRowsOptions Type Error for startIndex: Expects a string when it should expect a number
#### Environment details

  - OS: macOS 14.4.1
  - Node.js version: 20.11
  - npm version: 10.2.4
  - `@google-cloud/bigquery` version: 7.5.2

#### Steps to reproduce

  1. Try to pass `startIndex` as an integer option to `getRows`. eg.`table.getRows({ startIndex: 31241 })`.
  2. Observe typescript error. Expects a string when this should be a number.
  3. Passing a number in as "string" allows it to compile.

#### Expected behaviour
`startIndex` should work with an integer. As it is right now I need to convert the integer to a string eg. `31241.toString()`

#### Relevant fixes in the code
https://github.com/googleapis/nodejs-bigquery/blob/74aa1501452c36af7969bb4a46b996485d9ca91b/src/types.d.ts#L5903
https://github.com/googleapis/nodejs-bigquery/blob/74aa1501452c36af7969bb4a46b996485d9ca91b/src/types.d.ts#L6055

arjunmehta:
startIndex should be an 0-indexed integer. But as it is was expecting a string which is incorrect and caused a lot of confusion, preventing compilation and requiring odd workarounds.

- [ ] Ensure the tests and linter pass

Fixes googleapis/google-cloud-node#7230 🙏

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -5900,7 +5900,7 @@ declare namespace bigquery {
       /**
        * Zero-based index of the starting row.
        */
-      startIndex?: string;
+      startIndex?: number;
       /**
        * Optional: Specifies the maximum amount of time, in milliseconds, that the client is willing to wait for the query to complete. By default, this limit is 10 seconds (10,000 milliseconds). If the query is complete, the jobComplete field in the response is true. If the query has not yet completed, jobComplete is false. You can request a longer timeout period in the timeoutMs field. However, the call is not guaranteed to wait for the specified timeout; it typically returns after around 200 seconds (200,000 milliseconds), even if the query is not complete. If jobComplete is false, you can continue to wait for the query to complete by calling the getQueryResults method until the jobComplete field in the getQueryResults response is true.
        */
@@ -6052,7 +6052,7 @@ declare namespace bigquery {
       /**
        * Start row index of the table.
        */
-      startIndex?: string;
+      startIndex?: number;
     };
   }
 


alvarowolfx:
this file should not be manually modified, as it is generated from the Discovery doc (with the command `npm run types`).

arjunmehta:
@alvarowolfx Ah got it, I will close this PR then but the issue remains open. I am happy to contribute a solution, but not sure where to find/update the relevant Discovery Doc definition. If you want to point me to it I can attempt. Or someone who knows the procedure might be better suited to make the quick fix.

alvarowolfx:
Another thing to consider is that the `startIndex` attribute expects an `uint64` and that can't be represented with a integer in Javascript, which holds at most `2^53 - 1`. So users would have to use a `BigInt` and convert to `string` anyway.

PR #1372, opened 2024-06-04
ci: automate discovery types 

alvarowolfx:
Towards internal b/309836530 and supersedes #1366


.github/workflows/discovery.yaml
```diff
@@ -0,0 +1,24 @@
...
+  workflow_dispatch:    
+name: Update Discovery Generated Types
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+      - uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        with:
+          node-version: 16
```
leahecole, line 14: will these be managed by renovate or dependabot?

package.json
```diff
@@ -55,7 +56,9 @@
     "arrify": "^2.0.1",
     "big.js": "^6.0.0",
     "duplexify": "^4.0.0",
+    "execa": "^5.0.0",
     "extend": "^3.0.2",
+    "gaxios": "^6.0.3",
```
leahecole, line 61: should these be dev dependencies instead of regular ones?

scripts/submit-discovery-pr.js
```diff
@@ -0,0 +1,96 @@
...
+  console.log(foundChanges.join('\n'));
+
+  if (foundChanges.length === 0) {
+    console.log('No changes found');
+    return;
+  }
+
+  if (process.env.GITHUB_ACTIONS) {
+    await execa('git', ['config', 'user.email', 'yoshi-automation@google.com']);
+    await execa('git', ['config', 'user.name', 'Yoshi Automation']);
```
leahecole, line 43: @sofisl is this the most up to date git username and email we use for this kind of stuff?

.github/workflows/discovery.yaml
```diff
@@ -0,0 +1,24 @@
+on:
+  schedule:
+    - cron: '0 12 * * TUE'
+  pull_request: # REMOVE BEFORE MERGING
```
leahecole, line 4:
```suggestion
  pull_request: # REMOVE BEFORE MERGING
```

Can this be removed?

.github/workflows/discovery.yaml
alvarowolfx: yeah, renovate/dependabot managed those dependencies. See example here https://github.com/googleapis/google-api-nodejs-client/pull/3485

package.json
alvarowolfx: ah good catch, I forgot the `--dev` flag when I added then

.github/workflows/discovery.yaml
alvarowolfx: right now I only added it so we can trigger it here on this PR and to make sure it works. But will remove before merging

scripts/submit-discovery-pr.js
alvarowolfx: for context, I got from https://github.com/googleapis/google-api-nodejs-client/blob/main/src/generator/synth.ts

scripts/submit-discovery-pr.js
```diff
@@ -0,0 +1,93 @@
...
+  if (!githubToken) {
+    throw new Error('please include a GITHUB_TOKEN');
+  }
+
+  try {
+    // Open the pull request
+    await gaxios.request({
+      method: 'POST',
+      headers: {
+        Authorization: `token ${githubToken}`,
```
chingor13, line 64: This looks like it should be a bearer token
```suggestion
        Authorization: `Bearer ${githubToken}`,
```

sofisl (approved):

scripts/submit-discovery-pr.js
sofisl: yes!

PR #1446, opened 2025-03-06
chore: fix owlbot configs, regenerate

leahecole:
I worked with @sofisl extensively to get owlbot to properly run manually. Until we automate it, I will periodically run it to regenerate the library. This is the first of such refreshes. It includes 

- updates to .Owlbot.yaml
- the addition of a missing .Owlbot.lock.yaml
- a change to owlbot.py to exclude src/index.ts, which, if not excluded, will result in the library not being able to compile
- the addition of some missing rowAccessPolicy methods and other small updates to protos
- manual updates to bigquery_test.ts that
      - add tests for missing rowAccessPolicy methods
      - fixed up some sync tests that had some async leftovers
      - removed an erroneous todo
      - improve the error message for the check for # of methods to remind the maintainer to add tests for newly added methods instead of only updating that test
- adds a linkinator config file to skip two links that linkinator doesn't have access to (at the recommendation of @sofisl)
- manually fix broken links that were fixed in https://github.com/googleapis/gapic-generator-typescript/pull/1691


src/index.ts
```diff
@@ -12,11 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-// ** This file is automatically generated by gapic-generator-typescript. **
-// ** https://github.com/googleapis/gapic-generator-typescript **
+// ** This file is automatically generated by synthtool. **
+// ** https://github.com/googleapis/synthtool **
```
leahecole, line 16:
```suggestion
// ** https://github.com/googleapis/gapic-generator-typescript **
```

leahecole:
I also need to open a PR to synthtool

alvarowolfx (approved):

PR #1448, opened 2025-03-14
feat: don't overwrite handwritten methods

leahecole:
Address b/396701470


scripts/gen_uber_class.ts
```diff
@@ -357,6 +358,18 @@ function buildClientConstructor(clients) {
   return output;
 }
 
+// Looks at existing bigquery.ts - all handwritten code MUST be
+// in between // Begin handwritten and // End handwritten tags
+// or it will be overwritten
+async function readHandwrittenCode(): Promise<string> {
+  let handwrittenCode = '\n';
+  const input = await fsPromises.readFile('../src/bigquery.ts', 'utf8');
```
alvarowolfx, line 366: as we are reading the file on a synchronous way here, [fs.readSync](https://nodejs.org/api/fs.html#fsreadsyncfd-buffer-options) can be used and avoid making this function `async` and having to call `await` later.

alvarowolfx (approved):
minor thing on reading the file with promises vs using readSync.

scripts/gen_uber_class.ts
leahecole: right on - I was back and forth about if I wanted to use this, createReadStream, or readFile. Totally fine to do sync here given how small the file is. I'll make that change

PR #1461, opened 2025-04-09
autogen delete call handling

leahecole:
fixes b/396701485

This PR adds logic to the centralized client generator that does the following

- refactors some previously existing code into helper functions `optionsOrCallbackHelper`
- adds a new helper function `returnStatementHelper` that encompasses some existing logic and adds additional logic to handle delete calls
- adds logic to intercept errors from gax due to BQ's improperly formatted delete calls - this is partially done in `returnStatementHelper` (the callback case) and partially done by creating a new async function for each delete call - this async function means we can use async/await syntax and don't run into the awful type errors we would run into if we added async to the top level `delete*` functions (related to the optional callback and varying return types)

### DIFF
--- a/scripts/gen_uber_class.ts
+++ b/scripts/gen_uber_class.ts
@@ -55,6 +55,15 @@ const LICENSE = `
 
 `;
 
+const DELETE_CALL_COMMENT = 
+`        
+// BQ behaves unusually and doesn't return what we expect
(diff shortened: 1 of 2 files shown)


leahecole:
don't merge until we get approval on doc

leahecole:
Obsolete! :)

PR #1462, opened 2025-04-09
feat!: simplify bigquery.query method interface

### LINKED ISSUE #1400: Simplify `query` methods to return just `QueryRowsResponse`
For the [type returned by the query](https://github.com/googleapis/nodejs-bigquery/blob/efc36362175dc2f8035518db3dcb44b259429c7c/src/bigquery.ts#L2120) method, for historical/backward compatibility reasons, the method signature shows that it can return both a [`Promise<QueryRowsResponse>`](https://github.com/googleapis/nodejs-bigquery/blob/efc36362175dc2f8035518db3dcb44b259429c7c/src/bigquery.ts#L87) (which has 3 positional parameters) or [`Promise<SimpleQueryRowsResponse`](https://github.com/googleapis/nodejs-bigquery/blob/efc36362175dc2f8035518db3dcb44b259429c7c/src/bigquery.ts#L98) (which has only 2).

Ideally we should only return `QueryRowsResponse` and remove any reference to SimpleQueryRowsResponse. This is going to be a breaking change, so we should schedule this for the next major node upgrade release.

alvarowolfx:
Fixes #1400  🦕


.github/scripts/package.json
```diff
@@ -14,8 +14,8 @@
         "js-yaml": "^4.1.0"
     },
     "devDependencies": {
-        "@octokit/rest": "^21.0.0",
+        "@octokit/rest": "^19.0.0",
         "mocha": "^10.0.0",
-        "sinon": "^20.0.0"
```
leahecole, line 19: do these really need to be downgraded?

src/bigquery.ts
```diff
@@ -2200,25 +2191,25 @@ export class BigQuery extends Service {
...
         if (typeof query === 'object' && query.dryRun) {
-          (callback as SimpleQueryRowsCallback)(null, [], resp);
+          callback && callback(null, [], null, resp);
           return;
         }
         // The Job is important for the `queryAsStream_` method, so a new query
         // isn't created each time results are polled for.
         options = extend({job}, queryOpts, options);
-        job!.getQueryResults(options, callback as QueryRowsCallback);
+        job!.getQueryResults(options, callback);
```
leahecole, line 2204: we don't typecast here anymore because there's only one option for callback type, right?

src/bigquery.ts
```diff
@@ -2746,6 +2737,9 @@ export class BigQueryDatetime {
       time = time ? ' ' + time : '';
       value = `${y}-${m}-${d}${time}`;
     } else {
+      if (value.length > 100) {
+        throw new Error('BigQueryDatetime input too long');
+      }
```
leahecole, line 2742: More context needed! Why is this here? Is this related to the query stuff? If it needs to be in this PR, please edit the message to specify that it needs to be 100 characters or fewer and consider a link to our docs

leahecole (changes requested):
love to see this simplification. most comments are questions and my one meta question is:

we're changing the type and setting `nextQuery` to null throughout - will there ever be an instance where it would not be set to null?

.github/scripts/package.json
alvarowolfx: this branch need to be updated, octokit/rest was updated by dependabot last week

.github/workflows/issues-no-repro.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
       pull-requests: write
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v3
```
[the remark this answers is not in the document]
alvarowolfx: this branch need to be updated, this was updated by dependabot last week

src/bigquery.ts
```diff
@@ -2200,25 +2191,25 @@ export class BigQuery extends Service {
     if (!queryReq) {
       this.createQueryJob(query, (err, job, resp) => {
         if (err) {
-          (callback as SimpleQueryRowsCallback)(err, null, resp);
+          callback && callback(err, null, null, resp);
```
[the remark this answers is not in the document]
alvarowolfx: we were typecasting before because it could be a `SimpleRowsCallback` or `QueryRowsCallback`

src/bigquery.ts
alvarowolfx: this was to solve a security issue, I'll move that to a separated PR, sorry about that.

.github/scripts/package.json
alvarowolfx: oh, actually owlbot is rolling back those deps 🤔

alvarowolfx:
closing this as we are planning on a new query experience, so doesn't makes sense to break the signature now and cause some headache cc @leahecole

PR #1479, opened 2025-05-21
chore: periodic regen

leahecole:
This series of commits is part of a periodic regen and additionally ensures that the autogen branch is node 18 ready. Additionally, it changes the trailing comma strategy in the central client generation to be compatible with the changes in the tsconfig


leahecole:
gotta look into this - I think that the generation is conflicting with lint

alvarowolfx (approved):

PR #1480, opened 2025-05-27
feat: job creation mode GA

alvarowolfx:
Previously, use of the QUERY_PREVIEW_ENABLED environmental variable to control whether this client would attempt to execute queries without creating jobs. Now that this feature is GA, this client has been updated to expose a new client creation option `defaultJobCreationMode` to control this behavior. Also this can be overridden per query by the `Query` config attribute `jobCreationMode`.

This library will no longer use QUERY_PREVIEW_ENABLED to control this behavior, and the environment variable is ignored.

Towards internal b/418823080 🦕


shollyman (approved):

samples/queryShortMode.js
```diff
@@ -17,12 +17,13 @@
 function main() {
   // [START bigquery_query_shortmode]
```
shollyman, line 18: can you update this?  bigquery_query_job_optional

leahecole (approved):

samples/queryShortMode.js
alvarowolfx: done

PR #1490, opened 2025-06-27
feat: allow to skip wrapping data into custom types

alvarowolfx:
Draft PR on an idea to avoid the issue reported on googleapis/google-cloud-node#7229, where for some customers, having the custom types actually makes it harder to manipulate data. Naming is hard, not sure what is the best option name here.

Fixes googleapis/google-cloud-node#7229 
Fixes googleapis/google-cloud-node#7227

### DIFF
--- a/src/bigquery.ts
+++ b/src/bigquery.ts
@@ -121,6 +121,7 @@ export type Query = JobRequest<bigquery.IJobConfigurationQuery> & {
   pageToken?: string;
   wrapIntegers?: boolean | IntegerTypeCastOptions;
   parseJSON?: boolean;
+  skipWrapCustomTypes?: boolean;
   // Overrides default job creation mode set on the client.
   jobCreationMode?: JobCreationMode;
 };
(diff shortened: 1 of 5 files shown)


src/bigquery.ts
```diff
@@ -579,14 +582,16 @@ export class BigQuery extends Service {
...
+   * @param {boolean} options.parseJSON parse a 'JSON' field into a JSON object.
+   * @param {boolean} options.skipWrapCustomTypes skip converting values into BigQuery custom types.
    * @returns Fields using their matching names from the table's schema.
    */
   static mergeSchemaWithRows_(
     schema: TableSchema | TableField | undefined,
     rows: TableRow[],
     options: {
       wrapIntegers: boolean | IntegerTypeCastOptions;
+      skipWrapCustomTypes?: boolean;
```
leahecole, line 594: nit - can this be the last option? I know order isn't necessarily preserved but I feel like we should have these options in the order in which we have added them

leahecole: additional comment on this line - Any reason why we can't call this wrapCustomTypes? I know it would flip the boolean logic later on, but it would be in line with the wrapIntegers naming scheme

src/bigquery.ts
```diff
@@ -2462,14 +2479,18 @@ function convertSchemaFieldValue(
   value: any,
   options: {
     wrapIntegers: boolean | IntegerTypeCastOptions;
+    skipWrapCustomTypes?: boolean;
```
leahecole, line 2482: same note about order of options

leahecole (commented):
initial small thoughts

PR #1521, opened 2025-08-18
chore: Models samples

leahecole:
Adds samples for the models client and a samples test. PTAL especially at the create model sample and how I'm polling - I'm not sure if I should be polling with getQueryResults or with getModel - I can do either - getModel might feel a little less awkward, but we did getQueryResults in the old sample.


samples/models/createModel.js
```diff
@@ -0,0 +1,92 @@
...
+        },
+      },
+    };
+
+    // Run query to create a model
+    const [jobResponse] = await bigquery.insertJob(request);
+    const jobReference = jobResponse.jobReference;
+
+    const getQueryResultsRequest = {
+      projectId: projectId,
```
alvarowolfx, line 67: I have some suggestions on the pooling mechanism here. 
1. Remove `timeoutMs` and leave default value from the service side (I think it's 10 seconds). With the 2 min that is set up currently, the script is only gonna hang waiting for a single call and not polling happen in the end.
2. Use `node:timers/promise` setInterval to build a timed async iterator, so this truly runs a for loop with an interval. Right now with the `while` loop + `setTimeout`, it is creating multiple setTimeouts at the same time without blocking every iteration.
3. `resp.status` is not from the `jobs.getQueryResults` API, I think is from the `jobs.get`. We need to check for `resp.jobComplete` and it's a protobuf `BoolValue`, so the need for the `.value`.
```diff
diff --git a/samples/models/createModel.js b/samples/models/createModel.js
index 964492d..3220820 100644
--- a/samples/models/createModel.js
+++ b/samples/models/createModel.js
@@ -20,6 +20,10 @@ function main(
   modelId = 'my_model',
 ) {
   // [START bigquery_create_model]
+
+  // Import node:timers package for pooling mechanism.
+  const { setInterval } = require('node:timers/promises');
+
   // Import the Google Cloud client library
   const {BigQueryClient} = require('@google-cloud/bigquery');
   const bigquery = new BigQueryClient();
@@ -62,28 +66,24 @@ function main(
     // Run query to create a model
     const [jobResponse] = await bigquery.insertJob(request);
     const jobReference = jobResponse.jobReference;
-
     const getQueryResultsRequest = {
       projectId: projectId,
       jobId: jobReference.jobId,
       location: jobReference.location.value,
-      timeoutMs: {value: 120000},
     };
-    let [resp] = await bigquery.jobClient.getQueryResults(
-      getQueryResultsRequest,
-    );
+
     // poll the job status every 3 seconds until complete
-    while (resp.status === 'RUNNING') {
-      setTimeout(
-        ([resp] = await bigquery.jobClient.getQueryResults(
-          getQueryResultsRequest,
-        )),
-        3000,
+    for await (const t of setInterval(3000)) {
+      const [resp] = await bigquery.jobClient.getQueryResults(
+        getQueryResultsRequest,
       );
-    }
-    if (resp.errors.length !== 0) {
-      throw new Error('Something failed in model creation');
-    }
+      if (resp.errors.length !== 0) {
+        throw new Error('Something failed in model creation');
+      }
+      if (resp.jobComplete?.value) {
+        break;
+      }
+    } 
     console.log(`Model ${modelId} created.`);
   }
   createModel();
```

samples/models/createModel.js
```diff
@@ -0,0 +1,92 @@
...
+  async function createModel() {
+    // Creates a model named "my_model" in "my_dataset".
+
+    /**
+     * TODO(developer): Uncomment the following lines before running the sample
+     */
+    // const datasetId = "my_dataset";
+    // const modelId = "my_model";
+
+    const query = `CREATE OR REPLACE MODEL \`${projectId}.${datasetId}.${modelId}\`
```
alvarowolfx, line 36: should we use a lighter model like in the test ? This one takes a bit longer to create and reserve mode resources, not sure if is gonna become flaky with time in our CI/CD pipeline

samples/test/models.test.js
```diff
@@ -0,0 +1,216 @@
...
+        location: jobReference.location.value,
+        timeoutMs: {value: 120000},
+      };
+
+      // Wait for the query to finish
+      let [resp] = await bigquery.jobClient.getQueryResults(
+        getQueryResultsRequest,
+      );
+      // poll the job status every 3 seconds until complete
+      while (resp.status === 'RUNNING') {
```
alvarowolfx, line 127: need the same changes as in the `createModel` sample that I suggested.

alvarowolfx (changes requested):

alvarowolfx (approved):

PR #1527, opened 2025-09-02
fix: reuse auth across clients

leahecole:
- adds logic to not reinitialize auth every time based on feedback from @tswast 
- adds a line to the package.json to delete the generation script .js file when it's done (causes lint issues, not needed after generation is done)


scripts/gen_uber_class.ts
```diff
@@ -334,8 +334,13 @@ function buildClientConstructor(clients: string[]) {
...
   */`;
-  let constructorInitializers =
-    '\tconstructor(options?: BigQueryClientOptions, subClientOptions?: SubClientOptions){\n';
+  let constructorInitializers = `\tconstructor(options?: BigQueryClientOptions, subClientOptions?: SubClientOptions){
+        // if the user has passed in auth, use it, otherwise
+        // initialize auth with default so it gets passed to the sub clients
+        // and doesn't need to be re initialized each time
+        subClientOptions = subClientOptions || {};
+        subClientOptions.opts = subClientOptions.opts || {};
+        subClientOptions.opts.auth = subClientOptions.opts.auth || new GoogleAuth();\n\n`;
```
tswast, line 343: Does node require scopes or does it default to the cloud-platform scope?

alvarowolfx (dismissed):

scripts/gen_uber_class.ts
leahecole: Node does not require them, they can be passed optionally but also iiuc, defaultScopes are set as part of client initialization ([example](https://github.com/googleapis/nodejs-bigquery/blob/30f2457f5311e312eb14a428f0c076caf039dfd9/src/v2/dataset_service_client.ts#L178-L180)) - I am fuzzy on how this `staticMembers.scopes` gets populated (I should learn). In gax, this [gets initialized](https://github.com/googleapis/gax-nodejs/blob/1e447b3174fc069d1809d2fddae8dd7f90cbaf0d/gax/src/grpc.ts#L174) with certain options that are passed in - this [call to gax](https://github.com/googleapis/nodejs-bigquery/blob/30f2457f5311e312eb14a428f0c076caf039dfd9/src/v2/dataset_service_client.ts#L163) - I should probably update this to make sure that when we initialize a new client, we're passing in the options that would have  made their way to gax and that nothing is missing. 

Thoughts @alvarowolfx ?

alvarowolfx: yeah, Scopes are already resolved by calling each client `scopes` method. Ref: https://github.com/googleapis/nodejs-bigquery/blob/30f2457f5311e312eb14a428f0c076caf039dfd9/src/v2/dataset_service_client.ts#L359

So it already includes the `cloud-platform` scope.

tswast (approved):

alvarowolfx (approved):
nice, liked the way Auth with API keys was used to check if the auth client is being reused.

PR #1530, opened 2025-09-05
feat: run samples tests in grpc and rest

leahecole:
- updates all samples so they can take in the type of transport as a parameter
- exempts a couple of problematic tests adding comments with tracking issues
- runs tests in parallel so they go zoom zoom (well, as much as zoom zoom in node can go) 🏎️


leahecole:
I'm going to close and reopen this because I accidentally opened it against the wrong head at first and can't get header check to recover

alvarowolfx (approved):

PR #1533, opened 2025-09-09
feat: simplify client instantiation

alvarowolfx:
I think the ergonomics of having the constructor accept two parameter can be improved. Currently it works like this:
```
export type SubClientOptions = {
  opts?: ClientOptions;
  gaxInstance?: typeof gax | typeof gax.fallback;
};

export type BigQueryClientOptions = {
  datasetClient?: DatasetServiceClient;
  tableClient?: TableServiceClient;
  jobClient?: JobServiceClient;
  modelClient?: ModelServiceClient;
  routineClient?: RoutineServiceClient;
  rowAccessPolicyClient?: RowAccessPolicyServiceClient;
};

constructor(
    options?: BigQueryClientOptions,
    subClientOptions?: SubClientOptions,
  ) 
```

My proposal would be to merge the two and accept just one:
```
import {
  ClientOptions,
} from 'google-gax';

export type BigQueryClientOptions = ClientOptions & {
  datasetClient?: DatasetServiceClient;
  tableClient?: TableServiceClient;
  jobClient?: JobServiceClient;
  modelClient?: ModelServiceClient;
  routineClient?: RoutineServiceClient;
  rowAccessPolicyClient?: RowAccessPolicyServiceClient;

  gaxInstance?: typeof gax | typeof gax.fallback;
};

constructor(opts?: BigQueryClientOptions) {
```

Draft implementation to capture thoughts and discuss.


leahecole:
Accepting just one is probably not a bad idea- I wonder if we even need to merge them, or if we should just accept passing in clients? Users can already pass client options to those, so passing them as a separate parameter with the big client is possibly overkill

leahecole (dismissed):

leahecole (approved):

leahecole (approved):

PR #1534, opened 2025-09-09
chore: update list samples to use *Async method

leahecole:
Update list samples to use the preferred *Async methods
Update datasets test to look for a dataset within the first 100 (update, I removed that 100 limit because it caused problems in long-door-651)


samples/test/datasets.test.js
```diff
@@ -162,7 +162,7 @@ describe('Datasets', () => {
         'node datasets/listDatasets.js bigquery-public-data',
       );
       assert.match(output, /Datasets:/);
-      assert.match(output, new RegExp('usa_names'));
+      assert.match(output, new RegExp('census_bureau_usa'));
```
alvarowolfx, line 165: why this is changing ?

alvarowolfx (approved):
overall looks good to me, just got a bit confuse on the assertion on dataset name.

PR #1581, opened 2025-11-18
ci: should not fail on 403/429 docs links

alvarowolfx:
Linkinator now supports custom error handling on status codes. Since our docs ci step have been failing due to 403 and 429 errors, I moved then to be a warning.

### DIFF
--- a/linkinator.config.json
+++ b/linkinator.config.json
@@ -6,6 +6,10 @@
     "http://goo.gl/f2SXcb",
     "img.shields.io"
   ],
+  "statusCodes": {
+    "403": "warn",
+    "429": "warn"
+  },
   "silent": true,
   "concurrency": 10
 }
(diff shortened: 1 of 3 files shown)


linkinator.config.json
```diff
@@ -6,6 +6,10 @@
     "http://goo.gl/f2SXcb",
     "img.shields.io"
   ],
+  "statusCodes": {
+    "403": "warn",
+    "429": "warn"
```
miguelvelezsa: I think 429 shouldn't be an issue https://github.com/JustinBeckwith/linkinator/issues/681

alvarowolfx: good point. But ugh, linkinator v7 requires Node 20, I was testing locally with Node 20 and thought that was finally gonna solve this issue, but we are on Node 18 yet. 😢

PR #1596, opened 2026-01-26
feat: support high precision timestamp strings on getRows calls

danieljbruce:
## Description

This PR achieves the same objectives as https://github.com/googleapis/java-bigquery/pull/4010, but for Node as requested in [the planning sheet](https://docs.google.com/spreadsheets/d/1cHAw6cV8k_s40hw28C8STtlymu3MXEvTY6PaKgdgSGE/edit?resourcekey=0-hPyTawvew5ugN037XCfZXg&gid=0#gid=0). The idea is that for reading rows, the users should be able to access high precision values for timestamps if high precision values are being stored on the backend.

## Impact

This PR follows a test driven development approach against the getRows method.

Tests were written to evaluate what happens when getRows receives various input values for timestampOutputFormat and useInt64Timestamp. These tests revealed that not only are high precision values not being delivered to users for ISO8601_STRING return types, but also other bugs exist like calls hang on getRows calls that fail and conversion logic throughs errors for some calls that fetch rows. This PR fixes all the bugs and ensures the values with the right precision are delivered to users.

This chart details the before code changes / after code changes results with impact highlighted in green:

image

The highlighted green impact shows that conversion logic has been updated to avoid the  'cannot convert' errors and with this new logic changes are also applied to the BigQueryTimestamp class to maintain high precision on timestamp values returned to users.

## Testing

New system tests to capture all useInt64Timestamp/timestampOutputFormat combinations for getRows calls.

PR #1604, opened 2026-02-02
feat: Add high precision TIMESTAMP values for queries

danieljbruce:
## Description

This PR makes it so that all requests to the `/queries` endpoint support returning high precision results for the TIMESTAMP type. This ensures that users will always get the highest precision possible for TIMESTAMP typed values requested in queries.

https://github.com/googleapis/nodejs-bigquery/pull/1596 adds similar support for the `/data` REST endpoint which may be useful for reference.

## Impact

Allows users to fetch high precision timestamps for calls to the /queries endpoint.

## Testing

System tests were added to make sure no matter what formatOptions they provide, the values sent back by the server will be delivered to the user without losing any precision.

## Additional Information

Notice that the PR adds `timestampPrecision: 12` to all calls. We don't want to make adding this property optional because it would require an API change so that the user can opt in to timestampPrecision=12. It is better to have no API change and deliver TIMESTAMP values that have high precision values always because high precision is always desirable.


system-test/high-precision-query-tests.ts
```diff
@@ -0,0 +1,165 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as assert from 'assert';
+import {describe, it, before} from 'mocha';
+import {BigQuery} from '../src';
+
+describe.only('High Precision Query System Tests', () => {
```
alvarowolfx: why do we need a separated file ? The previous PR already deviated from the pattern of having system tests in the https://github.com/googleapis/nodejs-bigquery/blob/main/system-test/bigquery.ts and added https://github.com/googleapis/nodejs-bigquery/blob/main/system-test/timestamp_output_format.ts, can we at least bundle all timestamp related tests into the same file ?

And I'm going to bring again that we don't need the amount of combinations and system tests to check if the support for picosecond resolution is working. More tests doesn't mean more coverage necessarily.  Just with those two PR we are adding 28 query executions, where previously we had a total of around 50. Not saying that the coverage was perfect before, but adding too many system/integration tests makes CI slower, and I'm not seeing enough justification for the amount of combination and tests being added here.

system-test/high-precision-query-tests.ts
```diff
@@ -0,0 +1,165 @@
...
+  let bigquery: BigQuery;
+  const expectedTsValueMicroseconds = '2023-01-01T12:00:00.123456000Z';
+  const expectedTsValueNanoseconds = '2023-01-01T12:00:00.123456789123Z';
```
alvarowolfx: and this picosecond. Same applies to the previous https://github.com/googleapis/nodejs-bigquery/pull/1596/changes#diff-e82cb9428a68678cc305e12e59f2e21761622b9fcfcf45d4caab5e80657e2b9a test file

alvarowolfx (changes requested):

